PR #12, opened 2015-07-27
Use the provided included elements to render nodes

nlsrchtr:
This will remove the loading of `providers` for each Node if rendered in `NodesController#index` and therefor speed up loading time of Nodes from more than 450ms to under 150ms on my development machine.

Since there are no tests for this method included, would be great to know if this works as expected.


lennerd:
Nice catch and thank you for your help! :+1:

Issue #13, opened 2015-08-28
Widget implementation


skade:
* [ ] Frontend implementation
* [ ] Backend implementation

PR #16, opened 2015-09-16
Map widget

joustava:
#13 

- [x] widget models
- [x]  We can't use the authorization_token from backend because of missuse, we need some extra embed key for every user
- [x] The backend only needs a simplified map layout to serve the widget content (without header and footer)
- [x] The frontend should check for the provider id
- [x] The category filter in the frontend should only be visible if a query parameter for the category_filter is provided
- [x] Nodes should be loaded with the provider and embed key if available inside the frontend
- [x] Search as you type using Photon (komoot) component
- [x] Preview component
- [x] Form saving on changes. (mout/debounce immutable)
- [x] Open links inside the widget in a new window
- [x] Fix dropdowns in node popup
- [x] I18n component, Translations
- [x] Write tests
- [x] Fix dropdown
- [x] Refactore styling
- [x] Use boundingbox for search results
- [x] Min and max values for width and height
- [x] Fix profile form to formtastic
- [x] Reload widget only if certain properties got updated
- [x] Zoom level in embed code (if possible higher 16)
- [x] Explanation text under embed code
- [x] Fix default widget properties
- [x] Add translations
- [x] Link back to Wheelmap


lennerd:
I just added a first attempt to implement react into the rails asset pipeline. It's still failing in deployment (addressed in #30) but should work as expected in the local installation.

I would like to use this new implementation to start with our first react based frontend application – the embed widget. Please ask me if you have any questions to this. Some urls for the new gems:

https://github.com/browserify-rails/browserify-rails
https://github.com/reactjs/react-rails

This is the file where I tested the new gems:
https://github.com/sozialhelden/wheelmap/blob/map-widget/app/assets/javascripts/test.js

The `require` statement will be recognized by the extended asset pipeline so that browserify can wrap the files with specific code to enable component based loading. This way we will be able to use tools like [redux](https://github.com/rackt/redux) best available via npm and seperated into multiple smaller components we can select from.

Sadly there was not very much more possible today. Hope you have some frontend developer resources available to help out here.

lennerd:
To make this more clear:

For now I only want to use React for the widget generator, not to rebuild the whole homepage map.

lennerd:
@joustava Just checked our history. Why didn't you pull the last changes from git instead of merging changes from the remote to the local repository?

screenshot 2015-10-05 22 11 38

lennerd:
Please don't push any new changes. I need to clean up the map-widget branch so @christoph-buente can make a PR for a new important feature on a working master branch.

holgerd:
Load time of some elements are really slow, see screenshot (1.4min)

wheelmap_-_find_wheelchair_accessible_places

holgerd:
When I center the widget in "Paris" or "New York City" I do not see any POI pins. But I can see them in "Nürnberg" and "Berlin".


wheelmap_-_find_wheelchair_accessible_places


wheelmap_-_find_wheelchair_accessible_places

holgerd:
bug
- [ ]  I opened http://staging.wheelmap.org/en/profile/widgets and entered „Nürnberg“ in the „center“ input field. A dropdown opens with the text „Searching“ and nothing happens.

lennerd:
@skade I still have problems with deploying to staging as new assets seems to be not precompiled. It's either a problem with caching or the precompilation process. I'm not quite sure if I destroyed something somwhere. Can you take a look into this?

lennerd:
@skade I tried to add add the components assets to the precompile array. Without success. Can you take a look on this? When there are questions, please contact me via this issue, email or Slack.

Issue #24, opened 2015-09-28
Travisify


skade:
Get the test suite running on travisci.

lennerd:
Thanks to @christoph-buente, who fixed this.

christoph-buente:
I added a simple travis config and changed some more minor things to get travis rolling. Now there are still 10 specs failing.

christoph-buente:
I had to add some migrations to let the db/structure.sql reflect the current state of the schema. Now all specs are passing.

Unfortunately 4 of the cucumber tests are failing. Please have a look, if you can fix it.

PR #12 had to be rewritten as it largely broke the cucumber specs. But this is also done now.

I also removed the restriction to let travis only build the master. This means for future pull request travis should build them and give feedback to the contributor.

PR #33, opened 2015-10-05
Travisify

lennerd:
See #24


lennerd:
The cucumber tests are passing now. But the last task still aborts and results in failing travis build.

Any ideas?

Issue #20, opened 2015-09-25
Toilettenstatus aktualisieren in API


skade:
Vom Tim Freiheit:

Toilettenstatus akualisieren. (http://wheelmap.org/api/docs/resources/nodes)
Man kann über den Entpunkt "/api/nodes/{node_id}/update_wheelchair" den Status eines Nodes verändern. Ein Äquivalent gibt es für den Toilettenstatus allerdings nicht.
Im Idealfall für uns würde das über den selben Entpunkt laufen. Also beispielsweise mit 
"/api/nodes/{node_id}/update_wheelchair?wheelchair=yes&toilet=yes".

lennerd:
@timfreiheit Is this still active? cc @joustava

timfreiheit:
@lennerd yes

christoph-buente:
This should have solved with this commit: 
https://github.com/sozialhelden/wheelmap/commit/5a7345637cde2dc8dbdb6eb1176b9562c070ec2d

christoph-buente:
I just saw, that I updated the inline documentation with apipie, but I could not make it work in production. The old api documentation is still in place. I will provide an update for that.

https://rubygems.org/gems/apipie-rails

timfreiheit:
the endpoint is still not reachable:
```
http://staging.wheelmap.org/api/nodes/13/update_toilet?api_key=XXXXXX&wheelchair_toilet=yes
```
results in:
```json
{
  "error": "No route matches '/api/nodes/13/update_toilet', check API documentation for available resources: /api/docs/resources"
}
```

christoph-buente:
Hmm, the implementation is there, see

https://github.com/sozialhelden/wheelmap/blob/master/app/controllers/api/nodes_controller.rb#L184

But for some reason the routes seems to have gone. @lennerd can you check if it has been killed by a merge/rebase?

lennerd:
@timfreiheit I updated the current master to have an `update_toilet` url. Does that fix your problem?

timfreiheit:
@lennerd yes it looks good now. Thank you

lennerd:
:+1:

PR #43, opened 2015-10-14
Apidoc

christoph-buente:
* updated old style documentation to mention the update_toilet endpoint.

* generated a static version of the api documentation from inline comments in the https://github.com/sozialhelden/wheelmap/blob/master/app/controllers/api/nodes_controller.rb

It can be updated like this:

```bash
bundle exec rake apipie:static OUT=public/apidoc
```

Currently it only exists for the nodes controller. But i think it superior to the current API documentation, as it is nice to look at and the comments are close to the api code and should be ease to keep up to date.


christoph-buente:
This should also resolve #20

Issue #32, opened 2015-10-05
CSV-Export of Wheelmap-POIs should support >500 entries 


holgerd:
As admin I would like to be able to export more than 500 POIs as CSV-file. Right now only <500 entries are possible to be exported at http://wheelmap.org/en/admin/pois .

Example: download a CSV file of all POIs (yes, no, limited, unknown) in the category "food".

Requested by the wheelmap community manager.

joustava:
@holgerd, deployed a fix to staging. As far as I tested it it seems to download as many pois as the current filters returns. (so over 30000 pois without filters, which can take a while)

holgerd:
@joustava What is the maximum now?

I tried to download all 623.000 "unknown" places with this URL:
http://staging.wheelmap.org/en/admin/pois.csv?order=osm_id_desc&page=1&scope=unknown_accessibility

After a few seconds I got a "something went wrong" error.

It works with smaller sets such as the 5.900 fully accessible bakeries:
http://staging.wheelmap.org/en/admin/pois.csv?order=osm_id_desc&page=1&q%5Bnode_type_id_eq%5D=45&scope=fully_accessible&utf8=%E2%9C%93

christoph-buente:
Well, there is a timeout if course. So if rails takes a long time to render the csv information, nginx will show the error page at some point.

joustava:
So, it seems we are limited by the unicorn timeout, but capping the exported pois by an arbitrary number doesn't seem right to me either. Would it make sense to disable export features whenever the search result are to big to handle instead of limiting poi csv records to export?

Or can we configure unicorn to have a slightly higher timeout and make it support streaming, and generate the csv in batches?

christoph-buente:
Or maybe we can trigger a background job and put the file into the dropbox.

joustava:
Seems streaming this is supported out of the box in the newest version: https://github.com/activeadmin/activeadmin/pull/3038

christoph-buente:
Nice find! :+1:

holgerd:
@joustava The feature is fine as it is, please deploy and let me know when it is live. Just discussed it with Svenja and she said she will only download CSVs with less than 100.000 POIs, which already works. Thanks! :)

joustava:
@holgerd it is on staging for review.

holgerd:
@joustava thanks. The CSV-download stops after 1.4MB which is ca. 13.000 rows. I'll ask S. if that size is enough.

[...]

PR #45, opened 2015-10-21
Enable downloading large amounts of pois in CSV format

joustava:
Handles issue #32

Will download all Pois that are in the context of the current active filters.


christoph-buente:
I think it would not help comment out the max csv records, but put it to a very high number
```ruby
def max_csv_records
  10_000
end
```

joustava:
But if you put a very high number say 10000 and you want to download all 10001 that are returned by your filter you will miss one.

Xylakant:
There is actually a reasonable limit which would be what excel accepts as number of rows in a single spreadsheet. It used to be about 65000, now it's about 1.000.000 https://superuser.com/questions/366468/what-is-the-maximum-allowed-rows-in-a-microsoft-excel-xls-or-xlsx#366473

PR #53, opened 2015-11-04
Fix some typos

catoblepa:


holgerd:
thanks, we are using https://www.transifex.com to manage all translations and have implemented your typo corrections there. These are live already, just wanted to let you know!

Issue #71, opened 2015-12-01
Mysql2::Error: Table 'wheelmap_development.providers' doesn't exist 


1000miles:
### CASE:

This Myslq2 error happend a few times when trying to run rake commands. Even though the table for 'providers' is already created in the initial_schema, something in the migrations processes seems to always empty the database tables. I tried to drop, create & migrate a couple of times but yet the databases tables were empty (see step 2 below). To fix this, we have used a manual solution (see step 4) and for now it seems to work again. However, we wanted to document it in case this happens again. Maybe anyone can have a deeper look into this? 

### STEP 1:
rake db:migrate

## ERROR OUTPUT:
rake aborted!
*Mysql2::Error: Table 'wheelmap_development.providers' doesn't exist*
/Users/wheelmap/vendor/bundle/ruby/2.2.0/gems/activerecord-3.2.22/lib/active_record/connection_adapters/abstract_mysql_adapter.rb:243:in `query'
/Users/wheelmap/vendor/bundle/ruby/2.2.0/gems/activerecord-3.2.22/lib/active_record/connection_adapters/abstract_mysql_adapter.rb:243:in `execute'
/Users/wheelmap/vendor/bundle/ruby/2.2.0/gems/activerecord-3.2.22/lib/active_record/connection_adapters/mysql2_adapter.rb:213:in `execute'
/Users/wheelmap/vendor/bundle/ruby/2.2.0/gems/activerecord-mysql2spatial-adapter-0.4.3/lib/active_record/connection_adapters/mysql2spatial_adapter/main_adapter.rb:114:in `columns'
/Users/wheelmap/vendor/bundle/ruby/2.2.0/gems/activerecord-3.2.22/lib/active_record/connection_adapters/schema_cache.rb:12:in `block in initialize'
/Users//wheelmap/vendor/bundle/ruby/2.2.0/gems/activerecord-3.2.22/lib/active_record/model_schema.rb:229:in `yield'
/Users/wheelmap/vendor/bundle/ruby/2.2.0/gems/activerecord-3.2.22/lib/active_record/model_schema.rb:229:in `[]'
/Users//wheelmap/vendor/bundle/ruby/2.2.0/gems/activerecord-3.2.22/lib/active_record/model_schema.rb:229:in `columns'
/Users//wheelmap/vendor/bundle/ruby/2.2.0/gems/activerecord-3.2.22/lib/active_record/model_schema.rb:238:in `columns_hash'
[...]

PR #89, opened 2016-01-06
Fix errors on initial db migrations

schultyy:
PR for #71 

This is a workaround to ensure that users are able to run an initial migration on their database.
Before they failed because ActiveAdmin complained about missing tables which would have been created during migration.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -4,7 +4,13 @@ Wheelmap::Application.routes.draw do
 
   match '/ping' => 'ping#index'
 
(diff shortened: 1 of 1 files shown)


lennerd:
You should be able to merge this PR on your own. If the changes are coming from the team, the PRs are only for reviewing the changes at later point or for a better overview, but you can merge instantly if you're think everything is working fine.

lennerd:
Btw. thanks for the PR. :+1: :wink:

Issue #80, opened 2015-12-08
Counter on /profile not updating


holgerd:
A user reported after participating at #mapmyday on Dec 3, 2015:

> As part of the mapathon I tagged a good number of places, added 1 place and uploaded photos. The number of uploaded photos does show up, but the not the actual count of places tagged. Is this because the places are being held in a review queue? Can someone bring this up with Wheelmap?

Added screenshot:
![image](https://cloud.githubusercontent.com/assets/218506/11651668/d1cafff6-9d92-11e5-9b69-dbc244af0d93.png)

Holger had a look at the delayed_job queue:
![munin____wheelmap_org____spiderman_wheelmap_org____delayed_job](https://cloud.githubusercontent.com/assets/218506/11651669/d473cef4-9d92-11e5-94f6-7965235447dd.png)

## Slack conversation regarding this issue

@christoph-buente says:
> jobs are retried 25 times before giving up. (Some updates may not work, eg. if the places was remoded in OSM in the meantime)

@lennerd says:
> I also have a very low number of marked places on /profile. So does @holgerd

@Xylakant says:
> ich hab m ~ 16:20 ein paar fails im delayed_job-log, ansonsten reported der fleissig success

@christoph-buente 
> Probably counters are not correctly filled, can somebody please check this? Should be handled directly inside the jobs.

lennerd:
@holgerd Are you sure, that you marked all this places via the online map? Did you always login before you mark places?

holgerd:
I will test it again but the original bug reporters says he was not the only one who marked places without affecting the counter.

holgerd:
I marked a place and my counter went from 6 to 7 places immediately.

schultyy:
With improved logging implemented in #91 we understand better on what the delayed jobs actually work on.

holgerd:
@schultyy I guess can I close this?

schultyy:
Yes, this one can be closed

PR #91, opened 2016-01-11
Log more information about the current job

schultyy:
PR for #80 
This improves the output for the `UpdateTagsJob`. Additionally to the user id now also the element id and tags are logged.


schultyy:
@1000miles Can you check this PR?

Issue #87, opened 2015-12-23
amenity=community_centre spelled wrong (_center) in wheelmap categories


polarbearing:
The american spelling (amenity=community_center) of amenity=community_centre is deprecated but used in the categories to create a new entry. 

This has led to Germany having the highest concentration of the deprecated tag now :-(

Test example: http://www.openstreetmap.org/node/3906831259/history

lennerd:
We will definitely take a look into this. Thank you for reporting.

1000miles:
@lennerd Thank you. As soon as we have access  to the credentials, we will deploy and test it.

holgerd:
is this live now?

schultyy:
It's live now.

holgerd:
yay!

PR #94, opened 2016-01-13
Change community_center to community_centre

1000miles:
PR for #87 

@lennerd Can you pls take a look here, if we are on the right path? Anything else we need to definitely consider for our further research & work? Thank you.

cc @schultyy

### DIFF
--- a/app/views/nodes/_options_for_type.html.haml
+++ b/app/views/nodes/_options_for_type.html.haml
@@ -12,4 +12,3 @@
       - else
         %option{:label => t("poi.name.#{category.identifier}.#{node_type.identifier}"), :value => node_type.identifier }>
           =t("poi.name.#{category.identifier}.#{node_type.identifier}")
-    
--- a/config/locales/de/categories.yml
+++ b/config/locales/de/categories.yml
@@ -64,7 +64,7 @@ de:
         brothel: Bordell
         casino: Spielcasino
         cinema: Kino
-        community_center: "Stadthalle / Gemeindezentrum"
+        community_centre: "Stadthalle / Gemeindezentrum"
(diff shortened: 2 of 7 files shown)


lennerd:
Looks good to me. :+1:

lennerd:
Can you deploy to staging?

1000miles:
@lennerd thank you.

regarding your deployment question: not yet. can you pls give @schultyy and me access?

lennerd:
Please ask @Xylakant or @thermann78 for an access.

polarbearing:
Thanks a lot. Please let me know here once it is deployed, so we can start reviewing the current tags in OSM. 

One question: the commit adds this class:
```class ChangeCommunityCenterToCommunityCentre < ActiveRecord::Migration```

Does this change the spelling in the database once a user reviews the facility, or does it run automatically through the data? The latter would be a mechanical edit that is discouraged.

schultyy:
@polarbearing This change affects only newly created places with the node type community centre. We do not change already present data in OSM. For the places tagged with the old spelling we need to discuss how to proceed with those.

polarbearing:
My plan was to write a mail to talk-de once the bug is fixed in wheelmap, to ask local mappers for verification and manual migration. There are also new tags that can be considered manually for detailed description such as the type community_centre=* and for whom community_centre:for=* 

So, what does this migration class do, just for understanding?
```db/migrate/20160112153334_change_community_center_to_community_centre.rb```

schultyy:
This migration changes the spelling of community_center to community_centre in the wheelmap database. That value is then used when a user creates a new place via the wheelmap interface.

PR #109, opened 2016-01-25
Fix rspec deprecation warnings

1000miles:
This PR removes deprecation warnings & converts files to new rspec 3 syntax.


1000miles:
@schultyy Thanks. Fixed. :)

PR #116, opened 2016-01-27
Guard for rspec

1000miles:
This is a PR for bringing back all guard gems which we need for rspec-update.


1000miles:
Related to https://github.com/sozialhelden/wheelmap-privateissues/issues/14

PR #120, opened 2016-02-08
Update to rspec 3

1000miles:
This is the PR for RSpec 3.4 Update.


1000miles:
Because the current Travis build failure has to do with 3 rspec test cases @schultyy and me would like to merge this PR though. We will fix the test failures in another branch where we are going to switch from cucumber tests to rspec tests.

features/step_definitions/web_steps.rb
```diff
@@ -54,7 +54,8 @@ def with_scope(locator)
 end
 
 When /^(?:|I )follow "([^"]*)"$/ do |link|
-  click_link(link)
+  first(:link, link).click
+  # click_link(link)
```
schultyy: Is this line still needed?

1000miles: Thanks for the hint. I removed it. :)

PR #122, opened 2016-02-11
Rspec 3 helper upgrade

1000miles:
This PR adds the `rails_helper.rb` to the spec/ folder and requires it in every spec (since RSpec 3.x).


schultyy:
@1000miles Can you update this branch with the latest changes from master?

Issue #128, opened 2016-02-12
Remove inherited_resources


schultyy:
We need to remove inherited_resouces because it's not supported in Rails versions > 4.2.

schultyy:
Estimate: 5PT

schultyy:
We also need to take the API namespace into account. Those controllers are using inherited_resources as well.

1000miles:
Fixed in PRs #131, #139, #140 & #157.

Closed.

schultyy:
I have to reopen this one because we're not finally done here yet. There are still classes using `inherited_resources`. `inherited_resources` works with Rails 4.2 so we have some time until this gets due. When we want to do the upgrade to Rails 5, `inherited_resources` needs to go finally.

PR #131, opened 2016-02-15
Remove inherited resources from categories controller

schultyy:
PR for #128 
This is the first PR to remove `inherited_resources` from the application starting with the CategoriesController. This also delivers missing specs for that controller.


spec/controllers/categories_controller_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+    end
+
+    describe 'the response' do
+      it 'has categories' do
+        expect(json_response['categories'].length).to eq 1
+      end
+    end
+  end
+
+  context 'GET #show' do
```
1000miles: Why do you use "context" instead of "describe" here?

spec/controllers/categories_controller_spec.rb
```diff
@@ -0,0 +1,36 @@
+require 'spec_helper'
```
1000miles: We should `require 'rails_helper'`, or?

schultyy: Yes, good point. Will do :)

PR #140, opened 2016-02-15
Remove inherited resources from node types controller

schultyy:
This PR removes `inherited_resources` from the `NodeTypesController`

Related to #128

### DIFF
--- a/spec/controllers/node_types_controller_spec.rb
+++ b/spec/controllers/node_types_controller_spec.rb
@@ -2,7 +2,7 @@ require 'rails_helper'
 
 describe NodeTypesController do
   let!(:node_types) { FactoryGirl::create_list(:node_type, 5) }
-  describe '#index' do
+  describe 'GET index' do
     context 'without ids' do
       before do
         get :index, format: :json
@@ -38,7 +38,7 @@ describe NodeTypesController do
     end
   end
 
-  describe '#show' do
+  describe 'GET show/:id' do
     before do
       get :show, id: node_types.first.id, format: :json
     end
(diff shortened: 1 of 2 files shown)


spec/controllers/node_types_controller_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+      before do
+        get :index, format: :json
+      end
+
+      it 'returns 200 status code' do
+        expect(response).to have_http_status(:success)
+      end
+
+      it 'returns a list of node types' do
+        expect(json_response['node_types'].length).to be > 1
```
1000miles: When running `expect(json_response['node_types'].length).to be > 1` (line 15) in my tests, this raises an error while running `expect(json_response['node_types'].length).not_to be nil` let Rspec pass. What do you think?

schultyy: What kind of error does it raise?

1000miles: This when I run `bundle exec rspec spec/controllers/node_types_controller_spec.rb`:
```
Failures:

  1) NodeTypesController GET index without ids returns a list of node types
     Failure/Error: expect(json_response['node_types'].length).to be > 1

       expected: > 1
            got:   0
     # ./spec/controllers/node_types_controller_spec.rb:15:in `block (4 levels) in <top (required)>'

Finished in 0.37425 seconds (files took 6.9 seconds to load)
6 examples, 1 failure

Failed examples:

rspec ./spec/controllers/node_types_controller_spec.rb:14 # NodeTypesController GET index without ids returns a list of node types
```
Now I see, it is because `spec/fixtures/nodes_types.yml` is not seeded yet when only running the single file. Thanks for the hint.

PR #143, opened 2016-02-16
Add default settings for spec_helper

1000miles:
This PR adds the default settings of Rspec 3.4 for the file `spec_helper.rb` which is currently almost empty.


schultyy:
I am not sure if we need this because most of the things is already defined in `rails_helper` and `spec_helper` includes it.

1000miles:
Check, so if it's only default and we currently don't need it prepared, then we should close it. Thank you. :)

PR #146, opened 2016-02-17
Convert locale feature to spec

schultyy:
This PR converts the cucumber locale feature to rspec


schultyy:
Why is this one Work in Progress?

spec/features/locale_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+      end
+    end
+
+    describe "Fallback to default language when supported language is not available" do
+      let(:languages) { "be;pt" }
+
+      it "has locale de" do
+        expect(I18n.locale).to eq :de
+      end
+    end
```
1000miles: @schultyy 

**We wanted to improve the readability with this:**
```
  describe "Languages in wheelmap" do
    before do
      Capybara.current_session.driver.header("Accept-Language", SUPPORTED_LANGUAGES)
      visit '/map'
    end

    describe "sets preferred language" do
      let(:languages)  { "de-DE;de;en" }

      it "has locale :de" do
        expect(I18n.locale).to eq :de
      end
    end

    describe "has fallback for default language when supported language is not available" do
      let(:languages) { "be;pt" }

      it "has locale :de" do
        expect(I18n.locale).to eq :de
      end
    end
```
**So that the output would be this:**
```
  Languages in wheelmap
    sets preferred language
      has locale :de
    has fallback for default language when supported language is not available
      has locale :de
....

Finished in 0.93462 seconds (files took 7.35 seconds to load)
5 examples, 0 failures
```

PR #147, opened 2016-02-17
Fix asset host to be compatible with SSL protocol

### LINKED ISSUE #135: Ensure that the widget loads only protocol relative resources
Loading of plain-http resources is disabled for TLS-protected resources, so the widget itself must not load any resource from plain http when it's loaded embedded on an https-enabled site. Either protocol-relative URLs should be used or as an alternative, all resources should be loaded via https only.

lennerd:
Fixes #135


schultyy:
Looks good to me.

lennerd:
Testes on staging. Still without success. Looks like Rails does not recieve information about whether or not it's accessed via SSL.

@Xylakant @schultyy Any more ideas?

lennerd:
Another evidence for the server not passing the SSL information is that Rails also has the capabilities to use a simple URL as a string without protocol. This way, Rails adds the protocol to the URL itself. But this does not work either.

schultyy:
It just came into my mind, I guess the SSL Offloading happens before the request comes to the Rails application. This means Rails will never see any https requests.

lennerd:
Do you know if there are any HTTP headers which we can use to pass this information? cc @Xylakant

Xylakant:
I'll change the nginx config to pass on the protocol.
--
Felix Gilcher
Geschäftsführer

m. +49 172 840 88 28

asquera GmbH
Ohlauer Straße 43
D-10999 Berlin

AG Charlottenburg, HRB 140808 B
Geschäftsführung: Felix Gilcher, Florian Gilcher

PR #154, opened 2016-02-18
Fix for #148

### LINKED ISSUE #148: 500 error on wheelmap.org/api/categories/1/node_types
- I call http://wheelmap.org/api/categories/1/node_types?api_key=MYAPIKEY&per_page=6 in my browser where MYAPIKEY ist the key I got from my profile page
- I expect to see a result set.
- Instead, I get a 500 error. Other API request with my API key seem to work, for example http://wheelmap.org/api/nodes/3325161208?api_key=MYAPIKEY

lennerd:
This PR pins inherited_resources to 1.4.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -24,7 +24,7 @@ group :default do
   gem "routing-filter"
   gem 'newrelic_rpm'
   gem 'acts_as_api'
-  gem 'inherited_resources'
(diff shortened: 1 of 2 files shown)


lennerd:
Fix is on staging and `node_type` API is accessable again.

PR #169, opened 2016-03-16
Convert nodes feature to spec

1000miles:
This PR replaces the cucumber test `nodes.feature` from this folder: https://github.com/sozialhelden/wheelmap/blob/master/features/ with rspec.


spec/features/nodes_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+      provider = FactoryGirl.create(:provider, name: "Stiftung Gesundheit")
+      poi = FactoryGirl.create(:poi)
+      FactoryGirl.create(:provided_poi, wheelchair: "yes", url: "http://www.arzt-auskunft.de", poi: poi, provider: provider )
+      visit(node_path(poi))
+    end
+
+    # Scenario: I want to see a decent title on the poi detail page
+    describe "poi detail page" do
+      it "has a decent title" do
+          title_text = "Dieser Ort ist auf Wheelmap.org: name, Bar, Essen & Trinken, Berlin"
```
schultyy: Can you fix the indentation here?

spec/features/nodes_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+
+      # Scenario: I want to see a link to a landing page on the poi detail page
+      it "has a link to a landing page" do
+        expect(page).to have_link('Alle Orte vom Typ', href: "/map/berlin/bar/wheelchair/yes")
+      end
+
+      # Scenario: I want to see a back link to the root page
+      it "has a back link to the root page" do
+        click_link('zurück')
+        expect(page).to have_current_path(root_path)
```
schultyy: I would recommend to not click a link and then check the path but to check if a link with text "zurück" and href `root_path` exists. Currently you test more than you need.

1000miles: As discussed with Jan yesterday, we keep this method here because we want to validate 2 steps: click-link (UI) & being on the right path.

schultyy:
LGTM

PR #171, opened 2016-03-23
Updates routes in preparation to upgrade to Rails 4.0

justahero:
This PR prepares the routes for an upgrade from Rails 3.2 to Rails 4.0.

A few things changed from 3.2 to 4.0. For one, named routes cannot be stated multiple times (see `sign_in` and `sign_up`). This constraint may be relaxed in later version of Rails. The `match` statement in Rails 4.0 requires the `:via` argument and does not match implicitly to all HTTP actions. In Rails 3.2 there is already support for specifying actions, e.g. `via: [:get, :post]`. In Rails 4.0 there is also support to specify `via: :all` that handles all actions. Therefore all `match` statements are adjusted and work in both Rails 3.2 and 4.0 (Further info in [3.7 HTTP verb constraints](http://guides.rubyonrails.org/v3.2/routing.html))

Most of the routes are good candidates to have only support for `:get` with probably the exception of `match '/api' => 'api/api#index'`. It's very unlikely that external applications use something different from `:get`, but to not break something for these apps I chose to specify all HTTP actions.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -2,7 +2,7 @@ Wheelmap::Application.routes.draw do
 
   apipie
 
-  match '/ping' => 'ping#index'
+  get '/ping', to: 'ping#index'
 
   #
   # The if statement here is workaround to ensure that users are able to do an initial migration
@@ -28,7 +28,7 @@ Wheelmap::Application.routes.draw do
 
   scope "map" do
     root :to => 'home#index'
-    match '/:region_id/:node_type_id/wheelchair/:wheelchair' => 'landing_pages#index'
+    match '/:region_id/:node_type_id/wheelchair/:wheelchair' => 'landing_pages#index', via: [:get, :post, :put, :delete]
   end
 
   devise_for :users, :skip => [:sessions],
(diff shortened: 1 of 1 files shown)


1000miles:
@lennerd @thermann78 cc @holgerd 

As @justahero already mentioned, Rails 4 requires to specify the routing methods instead of using `match` that currently accepts all http-methods (:get, :post, :put, :delete) per default. Before we can merge this PR we need someone at wheelmap to check and confirm the methods that @justahero changed in the `routes.rb` file. 

@thermann78 Can you please check & confirm the blog part? :)

Thank you.

1000miles:
@justahero 

As bespoken with @thermann78 and @lennerd today, we can remove following paths
from the routes (which were broken anyways):
```
  get "/ziemlich-beste-freunde",  :to => redirect("http://blog.wheelmap.org/zbf")
  get "/goeslondon",              :to => redirect("http://blog.wheelmap.org/mitmachen/goes-london/")
  get "/goes-london",             :to => redirect("http://blog.wheelmap.org/mitmachen/goes-london/")
```
Can you please update the PR? The methods you suggested for the other routes are also accepted by @thermann78 and @lennerd.

1000miles:
@justahero Can you pls repeat your build for Travis? Travis still doesn't like phantomjs/poltergeist. :)

PR #172, opened 2016-03-24
Convert terms feature to spec

1000miles:
This PR replaces the current cucumber [terms.feature](https://github.com/sozialhelden/wheelmap/blob/master/features/terms.feature) with `rspec terms_spec.rb`.

Please note, in the last test case (line 44-53) of the file `terms.feature` this part `Then I should be on the last poi's edit page`  fails because the current app implementation redirects to the last poi detail page ` /nodes/17` instead of the last poi's edit page `/nodes/17/edit`.

We need to clarify, if the test expectation or the app redirect implementation needs to be corrected here.


spec/features/terms_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+    expect(page.current_path).to eq(terms_profile_path)
+  end
+
+  scenario "I must accept terms and privacy before editing nodes" do
+    click_link("Bearbeiten")
+    expect(page.current_path).to eq(terms_path)
+    check("user_terms")
+    check("user_privacy_policy")
+    click_on('Speichern')
+    expect(page.current_path).to eq(node_path(poi))
```
schultyy: If this test currently fails I would set it to pending with a comment why it fails.

1000miles: This `expect(page.current_path).to eq(node_path(poi))` does not fail, only if we follow the cucumber test expectation with `expect(page.current_path).to eq(edit_node_path(poi))`.

Thanks for the hint, I will put a note into the file.

schultyy:
Looks good to me. Ready to be merged :+1:

PR #178, opened 2016-04-07
Fixes issue with installing PhantomJS on TravisCI

justahero:
This PR fixes an issue with installing PhantomJS on the Travis CI server.

A few of the recents builds failed due to a problem with the download of PhantomJS when the ruby gem is used. The [phantomjs gem](https://github.com/colszowka/phantomjs-gem) downloads the package from the main host service (bitbucket). Most CI servers have fixed IP addresses and hit download limitations if a lot of builds download the same package from the same location in a short period. This also happens for TravisCI (and other CI services). For more detailed information on this issue see [this discussion](https://github.com/ariya/phantomjs/issues/13953).

TravisCI knows this issue and already has this package pre-installed in their environments (see [documentation](https://docs.travis-ci.com/user/gui-and-headless-browsers/#Using-PhantomJS)), making this gem as not required, at least on the CI server. My recommendation is to install PhantomJS separately on the local machine (see README) and not via the Gemfile. Using the prepackaged version on TravisCI avoids build problems.

* removes gem
* adds information on how to install PhantomJS on the development machine for testing

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -104,7 +104,6 @@ group :test do
 
   gem 'capybara', '~> 2.6', '>= 2.6.2' # on mac, you need sudo port install libffi
   gem 'poltergeist'
-  gem 'phantomjs', :require => 'phantomjs/poltergeist'
   gem 'database_cleaner'
 
   gem 'cucumber', :require => false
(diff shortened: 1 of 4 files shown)


schultyy:
LGTM

1000miles:
Thank you for the PR, @justahero! :)

PR #182, opened 2016-04-12
Remove old legacy ie.js 

lennerd:
The ie.js file was used to polyfill HTML 5 in IE 8 and older. But according to our analytics the IE 8 usage drops below 0.5 %.

I removed the file when I added the new node layout. This PR fixes old references to the polyfill.


lennerd:
Fixes #183.

1000miles:
Looks good to me so far. Thank you!

PR #192, opened 2016-04-18
Convert user feature to spec

1000miles:
This PR replaces the cucumber [user.feature](https://github.com/sozialhelden/wheelmap/blob/master/features/user.feature) but is mainly set on pending - waiting for the revision of the user signup/sign-in/recovery feature.


1000miles:
As discussed, we are going to merge this PR and touch it again at a later time.

Issue #54, opened 2015-11-04
Add/Update place types (key/value-pairs) = categories


holgerd:
Some place types are used in OpenStreetMap but can not be used in Wheelmap until now. Some have changed.
- [ ] Add the key/value types described in this public google spreadsheet with wheelmap-category and Icon (as linked in spreadsheet). @christoph-buente has done this several times before, there should be something already. Spreadsheet attached and here: 

https://docs.google.com/spreadsheets/d/12yuBoD8HNaC0sMH3JgLQlImz8NOMcaYgY0kRQpMDlSQ/edit?usp=sharing

[Neue Ortstypen für die Wheelmap.xlsx](https://github.com/sozialhelden/wheelmap/files/27384/Neue.Ortstypen.fur.die.Wheelmap.xlsx)

holgerd:
scraper info for wheelmap.org
https://www.dropbox.com/s/qlbti2ksgi4hoxj/wheelmapohnemap.html?dl=0

scraper info for wheelmap.org/map
https://www.dropbox.com/s/eb12nft1fol0yvq/wheelmapMITmap.html?dl=0

holgerd:
url to download the copied HTML of scraper info:
https://www.dropbox.com/sh/cmkqqbzn4732lgv/AAB2yz8DBsqzfAf-W7FVA7ria?dl=0

holgerd:
Related to this there is a suggestion from OSM users https://lists.openstreetmap.org/pipermail/talk-de/2015-November/112250.html

- [ ] Make it possible to NOT show places with this combination: leisure=swimming_pool + access=private

lennerd:
Everything is nearly prepared. There is one bigger step missing: the icon sprite file.

All our apps (iPhone, Android and Windows Phone) are using this file to show icons on the markers. If we add icons to this spirte, the position of the icons inside the sprite will change. As all the apps are propably are hardcoding the icon positions, it would break the icons representation.

So we need a smarter way to add new icons and also be sure that our mobile apps are capable of displaying them. Maybe we can add information to our API. Open for discussion this is.

christoph-buente:
From what i know: There is no sprite file, but a zip file with alle the markers in it.

lennerd:
Yes, but I'm not sure, if the Android and iPhone are still using them. The Windows Phone App is using the sprite file though.

christoph-buente:
What sprite? This is not delivered by the api anyways

lennerd:
[...]

PR #196, opened 2016-04-19
WIP: Add new node types

lennerd:


lennerd:
@1000miles I startet to dig out my stuff for creating the new node types. Unfortunately I lost some of my seeds while creating a new branch for this.

I'm not quite sure, how seeds are handled by the Rails backend. How do we get them into the existing system?

lennerd:
Related to #54

1000miles:
@lennerd 

Can you please check this file, if you already have updated the fixtures with the new node_types?
https://github.com/sozialhelden/wheelmap/blob/enhancement/new-node-types/spec/fixtures/node_types.yml

lennerd:
That was the problem. Thank you! 🙈 

@1000miles suggested that it would be better to split this PR into smaller ones to be able to revert the database easier if something went wrong. I'll go for that. I will leave this PR open to save the important bits for the new PRs. cc @holgerd 

We also will need to wait until https://github.com/sozialhelden/wheelmap-privateissues/issues/3 is fixed.

Issue #93, opened 2016-01-13
Widget not working on SSL pages


holgerd:
Buge report we got from somebody trying to use our widget:

====

Hi there,

We have problems implementing the Wheelmap widget on https://dev.handiscover.com
It worked well when running on localhost, but we can't manage to get it running on the https site. If we load the iframe by the original url (http), the browser refuse to load it because of mixed content. Replacing the url with https doesn't work either, as it is 304 redirected to the http version.
We have both tried to load the widget with JavaScript and generated it in the html source, but no luck.

The source code we use is like this:
<iframe src="http://wheelmap.org/sv/embed/FvjdVcWsokKopF62RSmy/?lat=43.4803174&amp;lon=-1.5625447&amp;zoom=15" border="0" height="440" width="100%"></iframe>

Best regards, Lukas Mattsson, Handiscover

Xylakant:
do we provide the widget via HTTPS? Plain HTTP Content is usually forbidden by the mixed content policies.

lennerd:
Yes, rights, we are aware of this problem. And it already happens with some of our users.

I have no big expierence in adding an HTTPS protocol. But I think it should be no big deel to add it to our servers. In the application should be no changes needed.

Xylakant:
there's some stuff that might need to be done if we want to provide a proper user experience. HTST headers, links for transactional emails might need adjusting, some redirects might need work. We need to ensure we're not embedding scripts and CSS from non-https sources. Cloudflare etc,

First of all we need a certificate, and we certainly need some testing.

lennerd:
Good point about the assets. As far as I know, the only external assets we use are some analytics and social badgets. But they all should work fine with HTTPS. We still need to check this though.

> First of all we need a certificate, and we certainly need some testing.

I think the best way would be to first add SSL only to the staging server (if possible). So we can check every page and redirect and give the team the chance to go to every corner of the page.

thermann78:
Cloudflare should handle SSL without any problems, it's just a matter of settings. Let me know if some parts are switched to HTTPS so I just do the setup change on CF.

holgerd:
Let's collect concrete tasks for this issue.

Xylakant:
SSL is enabled for stage using a self-signed cert. A quick test shows that we're loading resources from plain http which breaks the site. The following tasks have been opened to track progress on the ticket:

- [ ] Create a wildcard certificate for wheelmap.org [#132]
[...]

PR #197, opened 2016-04-19
Refactore SSL assets

lennerd:
Related to #93

### DIFF
--- a/config/environments/staging.rb
+++ b/config/environments/staging.rb
@@ -54,7 +54,7 @@ Wheelmap::Application.configure do
   config.serve_static_assets = false
 
   # Enable serving of images, stylesheets, and javascripts from an asset server
-  config.action_controller.asset_host = "http://staging.wheelmap.org"
+  config.action_controller.asset_host = "staging.wheelmap.org"
 
   # Disable delivery errors, bad email addresses will be ignored
   # config.action_mailer.raise_delivery_errors = false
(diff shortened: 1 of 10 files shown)


lennerd:
@Xylakant I deployed this PR on staging, but it still does not use the correct protocol.

Xylakant:
the config change was deployed on live only, let me fix stage.

Xylakant:
try again please.

lennerd:
Looks good, thank you! 👍

@1000miles Please review so we can merge this into master finally.

config/environments/staging.rb
```diff
@@ -54,7 +54,7 @@
   config.serve_static_assets = false
 
   # Enable serving of images, stylesheets, and javascripts from an asset server
-  config.action_controller.asset_host = "http://staging.wheelmap.org"
+  config.action_controller.asset_host = "staging.wheelmap.org"
```
1000miles: @lennerd I guess this looks good to me to remove the protocol. 

I have a question:
1. Do we use an ssl-encryption package that is valid for all pages on wheelmap.org or are we somehow limited to some pages that can not use `https`? 

If yes, only ssl - then we could use  `config.force_ssl = true`, or?

lennerd: I'm not quite sure. also about forcing to use SSL. @Xylakant @holgerd do you have information here?

1000miles: @lennerd thanks. okay, it might be another task. we can postpone this question to a later time. :)

1000miles:
@lennerd Thank you!

PR #203, opened 2016-04-25
Convert search feature to spec

1000miles:
This PR replaces the Cucumber [search.feature](https://github.com/sozialhelden/wheelmap/blob/master/features/search.feature) and removes all Cucumber [step definitions](https://github.com/sozialhelden/wheelmap/tree/master/features/step_definitions).


justahero:
So far looks good to me. If no cucumber feature exists anymore I would recommend removing the cucumber specific gems from the `Gemfile`, remove the file `lib/tasks/cucumber.rake` and check the configuration files where cucumber appears in, e.g. `database.yml`, `open_street_map.yml`, `Guardfile` etc.

This can also be done separately in another PR, as this requires a bit more testing due to changes in the configuration files.

1000miles:
Thanks for the review! :) Will create another branch for the subsequent tasks. :)

PR #220, opened 2016-05-09
Complete cucumber removal

1000miles:
This PR removes all files and codes related to cucumber.


config/initializers/carrierwave.rb
```diff
@@ -49,7 +49,7 @@ def effectively_resize_to_limit(width, height, format = 'jpg', quality = 70)
   end
 end
 
-if Rails.env.test? or Rails.env.cucumber?
+if Rails.env.test? or Rails.env.rspec?
```
justahero: It's better to remove the `Rails.env.rspec?` check, as RSpec is run in the `test` environment.

1000miles: Both work, with and without `RAILS.env.rspec?`so I will remove it. Thanks for the hint. :)

justahero:
LGTM

PR #221, opened 2016-05-10
Fix usage of lat lng query parameters in ember app.

### LINKED ISSUE #213: Widget creator cannot change city
User reports he cannot change the widget's city from Berlin to anything else. It always stays at Berlin. Svenja could replicate that behavior.

I could not replicate this systematically, but saw the error as well at least twice on my machine with different OSM-user IDs.

More testing needed.

lennerd:
This should fix not updating Wheelmap widget location inside the widget builder.

It's not the best solution and I found out that the Ember App trys to update the location of the map multiple times. This needs to be urgently refactored. See #23.

Fixes #213.

### DIFF
--- a/app/assets/javascripts/WidgetBuilder/Form.js
+++ b/app/assets/javascripts/WidgetBuilder/Form.js
@@ -66,7 +66,7 @@ class WidgetBuilderForm extends React.Component {
                 <I18n scope={'users.profile.widget.center'}/>
               </label>
               <span className="form-wrapper">
-                <Photon url={'http://photon.komoot.de/api/'} lang={I18n.locale} limit={10}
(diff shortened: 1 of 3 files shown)


lennerd:
Okay, I surrendered. The current Ember implementation is ugly and frustrating. It's extremely hard to debug and see where error happens.

@holgerd I just added the hack you suggested and reload the iframe every time the iframe source is changed. This leads to some flickering but also to a working preview. This whole thing can be simplified or even made interactive (so that the preview changes the embed code) with the next version I'm building with React right now. I add a comment as soon I was able to deploy the fix on staging.

schultyy:
LGTM

Issue #246, opened 2016-05-20
Mysql2::Error in Admin::AddressesController#index


1000miles:
# 5 

**Occurrences:**
- local
- `/admin/addresses`

## Mysql2::Error in Admin::AddressesController\#index

```
Table 'wheelmap_development.addresses' doesn't exist

```

`Rails.root: /home/vagrant/wheelmap`

 [Application Trace](http://192.168.33.10:3000/admin/addresses#) | [Framework Trace](http://192.168.33.10:3000/admin/addresses#) | [Full Trace](http://192.168.33.10:3000/admin/addresses#) 

```
lib/rack_i18n_locale_switcher.rb:32:in `call'

```

## Request

#  

**Parameters**: 

```
None

```

[Show session dump](http://192.168.33.10:3000/admin/addresses#)

[Show env dump](http://192.168.33.10:3000/admin/addresses#)

## Response

#  

**Headers**: 

```
None

```

schultyy:
Related to #71

1000miles:
Closed.

PR #247, opened 2016-05-23
Fix database setup infos

1000miles:
We had a similiar error before that a table doesn't exist, please see : #71 
This is because some migrations might have been removed in the past. 

As a fix for the issue #246 we decided to update the `README.md` file to easy up the `getting started` process after a database has been dropped and recreated.


schultyy:
LGTM 👍

Issue #244, opened 2016-05-20
[Rails 4.0] Update gem `activeadmin`


1000miles:
In the process of upgrading from Rails 3.2.x to Rails 4.0 we use this ticket as a parent task for all the issues/errors associated with this upgrade.

### Overview:

- [x] **1** NoMethodError in Admin::PhotosController#index (Fixed in 3574ede18d76005ea82655e8104d4221aa12375a)
- [x]  **2** NoMethodError in Admin/providers\#index (Fixed in 9a7aef1adb867ec215910c9eaa36a9655fe6e368)
  - [x]  **6** Formtastic::UnknownInputError in Admin/providers#upload_csv (Fixed in db051f238fb37a01a3dcdae700716027147f5b65 )
  - [x] NoMethodError in Admin/providers#new (Fixed in 34be7ee6f1ddcfa7f660591d8f3b3ba7a5ce7475) 
- [x]  **2** NoMethodError in Admin/provided_pois\#index (Fixed in 9a7aef1adb867ec215910c9eaa36a9655fe6e368)
  - [x]  **6** Formtastic::UnknownInputError in Admin/provided_pois#upload_csv (Fixed in db051f238fb37a01a3dcdae700716027147f5b65 )
  - [x] NoMethodError in Admin/provided_pois#new (Fixed in 34be7ee6f1ddcfa7f660591d8f3b3ba7a5ce7475)
- [x]  **3** NoMethodError in Admin/users\#new (PR #250)
- [x]  **4** LoadError in Admin/photos\#new (Fixed in 3574ede18d76005ea82655e8104d4221aa12375a)

1000miles:
**Occurrences:**
- `/admin/regions/netherlands/photos`

## NoMethodError in Admin::PhotosController#index

undefined method `find' for nil:NilClass

Rails.root: /home/vagrant/wheelmap
Application Trace | Framework Trace | Full Trace

app/admin/photos.rb:13:in `collection'
lib/rack_i18n_locale_switcher.rb:32:in `call'

Request

Parameters:

{"region_id"=>"netherlands"}

Show session dump

Show env dump
Response

Headers:

None

---

# 2

**Occurrences:**
- `/admin/providers`
- `/admin/provided_pois`

##  NoMethodError in Admin/providers\#index

[...]

PR #248, opened 2016-05-23
This removes the search_methods call from the poi model

### LINKED ISSUE #238: [Rails 4.0] Check `search_methods` & replace it if needed
**Background infos:**
* method `search_methods :region_id_eq` not available anymore

**In order to upgrade from 3.2.x to Rails 4.0 we need to do:**
* check where else this method is needed than at the 1 place
* implement an alternative in model Poi, e.g. `ransack`?

schultyy:
Related to #244 and fixes #238 
Right now filtering by region works out of the box with ransack without any additional configuration


1000miles:
Sounds good. Thanks!

PR #250, opened 2016-05-23
Fix NoMethodError in admin/users/new

schultyy:
Related to #244 
Fixes `NoMethodError` in `admin/users/new`


1000miles:
Looking gooood! Thx! 👍

Issue #251, opened 2016-05-23
[Rails 4.0] Undefined method `icon`in Providers & Provided_Pois


1000miles:
**Error output:**
```
undefined method `icon'
``` 
**Occurrences:**

```
/admin/providers
/admin/provided_pois/
```

**Before the active_admin update, following syntax worked:**

```ruby

action_item do
    link_to upload_csv_admin_provided_pois_path do
      icon( :'arrow_up') + "Upload CSV"
    end
  end
```

**To move forward with the active_admin upgrade, we removed the icon part temporarily. We will take a look into this later.**

```ruby

 action_item do
    link_to upload_csv_admin_provided_pois_path do
      "Upload CSV"
    end
  end
```

1000miles:
Related parent issue: #244, point 2

schultyy:
If we have time left we can do this within the Rails 4 Upgrade but can also be done afterwards.

PR #253, opened 2016-05-23
Fix providers & provided_pois errors in activeadmin

1000miles:
This PR fixes the issues from #244 related to `providers` and `provided_pois`.


schultyy:
For now the icon is removed but there's a ticket to bring it back: https://github.com/sozialhelden/wheelmap/issues/251

PR #258, opened 2016-05-24
Fix broken button calls

schultyy:
PR for #244 
This PR fixes the two broken admin forms `region` and `node_type`. Those contained calls to `buttons` but the `buttons` method has been replaced by `actions`.


1000miles:
That's great! Thanks!

PR #260, opened 2016-05-24
Fix removed ActiveAdmin methods

schultyy:
This PR fixes method calls which failed originally. We already fixed some of these in #259 but some changes got lost during the merge. 
They got lost because we needed to revert a previous PR containing some of these changes and apparently that revert went wrong.


schultyy:
@1000miles can you pleaase check?

1000miles:
Check! 🍡 Merci.

Issue #230, opened 2016-05-18
[Rails 4.0]: Use new syntax for `SimpleForm` in initializer (due to deprecation warning)


1000miles:
**In order to upgrade from Rails 3.2.x to Rails 4.0 we need to do following:**
- initializer: `SimpleForm.form_class` to `SimpleForm.default_form_class`, due deprecation

Important:
needs to be done **before** Rails upgrade

schultyy:
We can't do this before installing Rails 4 because there's no compatible version of SimpleForm with the new `default_form_class` for Rails 3.

PR #264, opened 2016-05-25
Fix SimpleForm deprecation warning in initializer

schultyy:
PR for #230

This fixes the SimpleForm deprecation warning in the initializer.


1000miles:
Thanks! 🍰

PR #265, opened 2016-05-25
Fix encoding in NodesController spec

schultyy:
This PR fixes two broken tests where umlauts where represented as unicode sequences. Now the assertions just contain the regular umlauts.


1000miles:
Noiiiice! 👯

Issue #239, opened 2016-05-19
[Rails 4.0]: ActiveRecord::StatementInvalid, Mysql2::Error  - Cannot get geometry object from data


1000miles:
**In the process of upgrading Rails 3.2.x to Rails 4.0 we need to take a look at following error output:**
```
ActiveRecord::StatementInvalid, Mysql2::Error - Cannot get geometry object from data you send to the GEOMETRY field
```

schultyy:
```
  55) PlanetReader should import a new POI
      Failure/Error: Poi.import @columns, @to_be_created, :validate => false, :on_duplicate_key_update => @columns_without_create, :timestamps => false

      ActiveRecord::StatementInvalid:
        Mysql2::Error: Cannot get geometry object from data you send to the GEOMETRY field: INSERT INTO `pois` (`osm_id`,`version`,`tags`,`geom`,`status`,`created_at`,`updated_at`,`node_type_id`,`region_id`,`toilet`) VALUES (26735750,7,'---\naddr:city: Berlin\naddr:country: DE\naddr:postcode: \'10623\'\namenity: restaurant\ncuisine: spanish\nname: Mar & Sol\noperator: Adolf Präg GmbH & Co. KG\nwheelchair: \'yes\'\n','POINT(13.3228529 52.5065963)',1,'2016-05-25 08:59:50','2016-05-25 08:59:50',1,NULL,NULL) ON DUPLICATE KEY UPDATE `pois`.`osm_id`=VALUES(`osm_id`),`pois`.`version`=VALUES(`version`),`pois`.`tags`=VALUES(`tags`),`pois`.`geom`=VALUES(`geom`),`pois`.`status`=VALUES(`status`),`pois`.`updated_at`=VALUES(`updated_at`),`pois`.`node_type_id`=VALUES(`node_type_id`),`pois`.`region_id`=VALUES(`region_id`),`pois`.`toilet`=VALUES(`toilet`)
      # ./app/importer/planet_reader.rb:69:in `create_pois'
      # ./app/importer/planet_reader.rb:55:in `load'
      # ./spec/importer/planet_reader_spec.rb:60:in `block (2 levels) in <top (required)>'
      # ------------------
      # --- Caused by: ---
      # Mysql2::Error:
      #   Cannot get geometry object from data you send to the GEOMETRY field
      #   ./app/importer/planet_reader.rb:69:in `create_pois'
```

1000miles:
Fixed in PR #267.

PR #267, opened 2016-05-25
This fixes the mysql insert error for POI import

schultyy:
PR for #239 

This fixes the INSERT Error in the POI Import tests.


1000miles:
Merci for the PR. 🎯

Issue #262, opened 2016-05-24
[Rails 4.0] RuntimeError in Nodes#show on detail page


1000miles:
**Occurrences:**

- on node detail pages > click on button (Details)

# RuntimeError in Nodes#show

Showing /home/vagrant/wheelmap/app/views/nodes/_node_share.html.haml where line #13 raised:

`mailme_to: Use native mail_to method with Rails 4.0 and later.`

Extracted source (around line #13):

```ruby

         =link_to "//www.facebook.com/sharer/sharer.php?s=100&p[url]=#{node_url(node_share)}", :'data-toggle' => 'share', :'data-name' => 'facebook', target: '_blank' do
           =fa_icon('facebook fixed-width', text: t('nodes.show.share.facebook'))
       %li
         =mailme_to(nil, body: t('nodes.show.share.text', name: node_share.name) + " (#{node_url(node_share)})") do
           =fa_icon('envelope-alt fixed-width', text: t('nodes.show.share.email'))
```
Trace of template inclusion: app/views/nodes/_node_edit.html.haml, app/views/nodes/show.html.haml

Rails.root: /home/vagrant/wheelmap
Application Trace | Framework Trace | Full Trace

app/helpers/application_helper.rb:156:in `mailme_to'
app/views/nodes/_node_share.html.haml:13:in `_app_views_nodes__node_share_html_haml'
app/views/nodes/_node_edit.html.haml:8:in `_app_views_nodes__node_edit_html_haml'
app/views/nodes/show.html.haml:16:in `_app_views_nodes_show_html_haml
app/controllers/nodes_controller.rb:75:in `show'
lib/rack_i18n_locale_switcher.rb:32:in `call'

Request

Parameters:

{"id"=>"xxxx"}

Toggle session dump
Toggle env dump
Response

Headers:

None

1000miles:
Fixed in PR #268.

PR #268, opened 2016-05-25
Use native mail_to method in view template

schultyy:
PR for #262 

Fixes the Share via Email button on the node detail page.


1000miles:
Gracias! 💎

Issue #242, opened 2016-05-19
[Rails 4.0]: Fix i18n locale warnings


1000miles:
**In the process of upgrading from Rails 3.2.x to Rails 4.0 we need to do**
- check & fix i18n locale warnings

**Failure/Error Example**
```
Failure/Error: I18n.locale = session["locale"] - # ./lib/rack_i18n_locale_switcher.rb:30:in call' # ./spec/features/create_spec.rb:16:inblock (3 levels) in '
```

schultyy:
We need to figure out when these failures happen. On my local machine they do not occur, Travis don't report them either right now.

schultyy:
My comment above is obsolete because this behavior only occurs with Rails 4.

PR #270, opened 2016-05-26
Fix invalid locale error

schultyy:
PR for #242 

We queue this initializer after every other initializer + Rails because Rails itself sets `I18n.enforce_available_locales` to true where we require it be `false`.
The code in `lib/rack_i18n_locale_switcher.rb` depends on `I18n.enforce_available_locales` to be `false`.

This fixes the problem for now but is not optimal yet. We need to figure out if we actually need the `rack_i18n_locale_switcher.rb` generally.


justahero:
You could check if setting the option `I18n.enforce_available_locales ` in the file `config/application.rb` avoids the logic with the after hook.

schultyy:
@justahero tried it but doesn't work.

schultyy:
Deployed it on staging and tested it. Language switching works so far.

1000miles:
Works on my machine too and looks good to me. 
Thank you! 🍔

PR #271, opened 2016-05-30
Use PATCH instead of PUT

schultyy:
**Note: Cannot be merged before #270 is merged**
Use the `PATCH` verb instead of `PUT` for `/profile/terms`


1000miles:
Thanks! 🍦

PR #272, opened 2016-05-30
Fix asset pipeline for rails4

1000miles:
Since Rails 4.0 the asset pipeline for precompiling has changed e.g. from `/public/stylesheets` to `public/assets/stylesheets`, therefore we need to adjust some configs. 

This PR fixes the asset precompiling process with a new asset pipeline process, it:
- installs the gem `sprocket-rails` to use the benefit of the new asset pipeline
- moves some configs from `application.rb` to a separate initializer called `assets.rb`
- renames the empty application.css file to application.scss to use sass syntax
- imports required asset precompiling gems in the manifests
- adds the asset.paths for `app/assets/javascript`, `app/assets/images` and `app/assets/stylesheets`


schultyy:
Why are some settings moved into an initializer?

schultyy:
Concerning the initializer: Apparently it's the way how it's done in Rails 4+. But I couldn't find out why yet.

config/environments/production.rb
```diff
@@ -15,11 +15,7 @@
 
 
   # Compress JavaScripts and CSS
-  config.assets.compress = true
-
-  # Compress JS and CSS faster in production
-  config.assets.css_compressor = :yui
-  config.assets.js_compressor = :uglifier
+  config.assets.compress = false
```
1000miles: Why do you set `config.assets.compress` to false? As far as I know we should keep compressing for production (as it was set to true before) but what we can do is to disable :yui and :uglifier for now and find alternatives later. :)

schultyy:
We can remove the `assets` group from the Gemfile since Rails 4 doesn't support it anymore: 
https://stackoverflow.com/questions/16406204/why-did-rails4-drop-support-for-assets-group-in-the-gemfile

config/environments/production.rb
```diff
@@ -17,11 +17,16 @@
   # Compress JavaScripts and CSS
   config.assets.compress = true
 
+  # @TODO Enable compressors and find alternative for :yui
+  # Compress CSS and JS faster in production
+  # config.assets.css_compressor = :yui
+  # config.assets.js_compressor = :uglifier
+
   # Don't fallback to assets pipeline if a precompiled asset is missed
-  config.assets.compile = false
```
schultyy: This should still be `false`

config/environments/production.rb
```diff
@@ -17,11 +17,16 @@
...
+  config.assets.compile = true
 
   # Generate digests for assets URLs
-  config.assets.digest = true
+  config.assets.digest = false
```
schultyy: This needs to be `true`

schultyy:
Tested deployment on staging. Ready to be merged 💃

PR #273, opened 2016-06-02
remove failing photos controller test

schultyy:
We remove this test because it tests Ruby on Rails behavior here. The
test starts failing with Rails 4.x where it passed before with Rails
3.2.x. Rails by itself doesn't allow to send a POST request to a route
which accepts GET only.


1000miles:
Makes sense to remove it. Thanks! :neckbeard:

PR #275, opened 2016-06-02
Fix first batch of deprecation warnings

schultyy:
This PR fixes the first deprecation warnings.

- It removes `config.whiny_nils`
- It removes `config.active_record.auto_explain_threshold_in_seconds`
- It sets `config.eager_load` for all environments


1000miles:
Thanks for the PR! :octocat:

PR #276, opened 2016-06-03
Clean up deprecation warnings in Poi scopes

schultyy:
This fixes almost all warnings in `app/models/poi.rb`. There are two to three warnings left which need to be fixed in a follow-up pull request.


1000miles:
Thank you! 🌴

Issue #236, opened 2016-05-18
[Rails 4.0] Fix all deprecation warnings related to Rails 4.0 upgrade


1000miles:
**To upgrade from Rails 3.2.x to Rails 4.0 we need to do following:**

* fix all deprecation warnings related to the upgrade, currently counting: 

  ```
   63 (incl. duplicates)
   18 (netto)
  ```

schultyy:
I assign this to Thuy and me to split up the work of fixing the warnings.

1000miles:
**Last Update: 2016, June 3**

Deprecation warnings to be fixed:
- [ ] The Active Record auto explain feature has been removed. 
- [ ] require "activeresource" is deprecated and will be removed in Rails 3. Use require "active_resource" instead.
- [ ] Support for Rails < 4.1.0 will be dropped.
- [ ] config.whiny_nils option is deprecated and no longer works.
- [ ] Using #scope without passing a callable object is deprecated.
- [ ] Calling #scope or #default_scope with a hash is deprecated.
- [ ] Poi.belongs_to :node_type declaration are deprecated: :include.
- [ ] Calling #scope or #default_scope with a hash is deprecated. 
- [ ] Region.has_many :children declaration are deprecated: :order.
- [ ] Active Admin: using `action_item` without a name is deprecated!
- [ ] You didn't set config.secret_key_base.
- [ ] #connection is deprecated in favour of accessing it via the class.
- [ ] Model.scoped is deprecated. Please use Model.all instead.
- [ ] This dynamic method is deprecated. Please use e.g. Post.find_or_create_by(name: 'foo') instead.
- [ ] Relation#all is deprecated. If you want to eager-load a relation, you can call #load

PR #279, opened 2016-06-06
Fix action item deprecation warnings

1000miles:
This PR fixes the deprecation warnings on `action_item`. **Please merge #278 first.**
Related Issue: #236


app/admin/provider.rb
```diff
@@ -4,10 +4,12 @@
...
-  action_item do
-    link_to upload_csv_admin_provided_pois_path do
-      "Upload CSV"
-    end
+  action_item(:upload_csv, :only => :index) do
+    link_to "Upload CSV", upload_csv_admin_provided_pois_path
+  end
+
+  collection_action :upload_csv, title: "Upload CSV" do
+    render "/admin/provided_pois/upload_csv"
```
schultyy: What does `collection_action` do here?

schultyy:
Why are also changes in this PR which are not related to ActiveAdmin `action_item`?

1000miles:
This was a separate test on `provided_poi`. Thanks for the hint. I will remove the `collection_action` part.

schultyy:
There are still changes which do not belong to ActiveAdmin:
- `app/models/poi.rb`
- `app/models/provided_poi.rb`
- `app/models/user.rb`

PR #281, opened 2016-06-06
configure example secret_token in .env file

schultyy:
PR for #236 
This fixes the warning that `secret_key_base` hasn't been set


1000miles:
I read that in Rails 4.x the naming has changed from `config.secret_token` to `config.secret_key_base`. 
https://github.com/sozialhelden/wheelmap/blob/master/config/initializers/secret_token.rb#L7

Is it something we should consider, too?

schultyy:
No `secrets.yaml` is not supported in Rails 4.0. This feature is available at Rails 4.1 and later.

1000miles:
Merci for the PR. 🐎

PR #283, opened 2016-06-06
Fix warning `Region.has_many :children declaration are deprecated: :order`

schultyy:
PR for #236 

This warning is fixed by using a newer version of the `awesome_nested_set` gem.


1000miles:
Thank you! 🐰

PR #290, opened 2016-06-09
Rename stylesheets from `css.scss` to `.scss` due to deprecation warning

1000miles:
This PR fixes the deprecation warning `Extra .css in SCSS file is unnecessary. Rename.... `  
It is not directly caused by the Rails 4.x upgrade but when we updated the sass-rails version we had to consider this, too.

Related issue to read more: #236


1000miles:
The precompiling and the tests have passed when we renamed the stylesheets of the folder `relaunch` under Rails 3.2 **and** Rails 4.0, so we decided to close this PR and create a new one that can be merged directly into the master branch.
The reason for this is that the `Rails 4.0 upgrade` branch is growing and still in progress but others (e.g. @lennerd) can work with the renamed stylesheets meanwhile to be prepared for the change.

lennerd:
@1000miles Very good thinking! Thank you!

Issue #274, opened 2016-06-02
[Debugging] Add SimpleCov and Pry-Byebug


1000miles:
We want to improve our debugging process by adding two 2 gems:
- [x] [SimpleCov](https://rubygems.org/gems/simplecov) to check if specific lines are (still) used
- [x] Pry-ByeBug

1000miles:
Related PR for `pry-byebug`: https://github.com/sozialhelden/wheelmap/pull/292

1000miles:
Related PR for `SimpleCov`: #305

PR #292, opened 2016-06-10
Add pry-byebug for better debugging

1000miles:
This PR adds the gem `pry-byebug` in addition to `pry`.


justahero:
Resolves part of #274.

PR #293, opened 2016-06-10
Fix broken photos api

schultyy:
PR for #236 

This fixes the broken `asset_paths` tests. Since `asset_paths` is not present anymore in Rails 4 we use `asset_path` with the `:protocol => :https` option to generate https urls by default. `asset_url` does not work here because it tries to obtain the protocol from the current request by default. But since it uses `asset_path` under the hood we circumvent it and use `asset_path` directly.
To make the tests pass we need to have `asset_host` to be set in the test environment as well.


schultyy:
As discussed, I add tests which ensure that photos are accessible by the returned urls.

1000miles:
All tests have passed on my side, too. Thanks!

PR #299, opened 2016-06-14
Use browserify-gem and remove gulp workaround.

lennerd:


app/assets/javascripts/test.js
```diff
@@ -0,0 +1,9 @@
+const bla = 'Test';
+
+class Test {
+
+}
```
schultyy: As discussed, this can be removed

Gemfile
```diff
@@ -86,6 +86,8 @@ group :default do
   gem 'compass-colors'
   gem 'compass-rails'
   gem 'sprockets-rails', :require => 'sprockets/railtie'
+  gem 'sprockets-es6'
+  gem 'browserify-rails'
```
schultyy: can you pin the version numbers here for `sprockets-es6` and `browserify-rails`?

lennerd:
@schultyy Thanks for the feedback! Squased all commits and added changes based on feedback.

schultyy:
LGTM

PR #304, opened 2016-06-14
Remove unused ember from to be precompiled list.

lennerd:
Ember is never used standalone but always required from manifest files. This PR removes `ember.js` and `ember-data.js` from the precompile list, which further improves our asset precompile on deployment.


schultyy:
LGTM

1000miles:
Thanks for the PR! 🍊

PR #305, opened 2016-06-15
add simplecov for development and testing

schultyy:
PR for #274 

add `simplecov` for development and testing.


spec/rails_helper.rb
```diff
@@ -1,5 +1,7 @@
   # This file is copied to spec/ when you run 'rails generate rspec:install'
   ENV["RAILS_ENV"] ||= 'test'
+  require 'simplecov'
+  SimpleCov.start
```
1000miles: Do we want to use SimpleCov only for testings but not for the whole rails app?

Issue #282, opened 2016-06-06
Fix `community_centre` error (de.poi.name.leisure.community_centre) on staging


1000miles:
On the staging server currently the `community_centre` file is broken. We have to do another import of translation files to check if that happened during the last import or if that is because of other reasons. 

**Occurrences:**
- http://staging.wheelmap.org/node_types

Error:
```
translation missing: de.poi.name.leisure.community_centre
```

1000miles:
After our research we found out following:

- the default locale for the wheelmap app is set to german, means if a translation for a word (key) is not available, the language for the app falls back to german 
- on transifex english is the source language for all translations, means it uses the english [categories.yml](https://github.com/sozialhelden/wheelmap/blob/master/config/locales/en/categories.yml#L68 ) file but currently an older version which still contains `community_center` instead of `community_centre`
- not all categories.yml of different languages listed in the [locales](https://github.com/sozialhelden/wheelmap/tree/master/config/locales) folder contain the key, neither `community_center` nor `community_centre` which is why they are missing the poi translation anyways 

To do:
- the current german [categories.yml](https://github.com/sozialhelden/wheelmap/blob/master/config/locales/de/categories.yml#L73) needs to be corrected to `community_centre` in order to tell the wheelmap app that the key is existent  => To be done by: @1000miles
- the english categories.yml file that is uploaded on transifex needs to be exchanged with the newer version in order to be the reference for translations  => To be done by: @holgerd?
- make sure to always have the same keys in all locale files that need translations (even if the translations are not there yet) => To be done by: ?? (@holger can you please check with the team that was managing the translations so far?)

1000miles:
Fixed in PR #312

PR #306, opened 2016-06-16
Fix missing key by changing from `community_center` to `community_centre`

1000miles:
This PR fixes the error `translation missing: de.poi.name.leisure.community_centre` on:

- staging.wheelmap.org/node_types
- wheelmap.org/node_types

Fixes partly issue: #282

### DIFF
--- a/config/locales/de/categories.yml
+++ b/config/locales/de/categories.yml
@@ -70,7 +70,7 @@ de:
         brothel: Bordell
         casino: Spielcasino
         cinema: Kino
-        community_center: "Stadthalle / Gemeindezentrum"
+        community_centre: "Stadthalle / Gemeindezentrum"
         gallery: Galerie
         massage: "Massage-Studio"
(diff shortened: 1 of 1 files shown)


schultyy:
Is the german translation file the only one with the old key?

schultyy:
There's at least one more file with the old name: https://github.com/sozialhelden/wheelmap/blob/9f4bbe595573251d9b136ab5ce8c19e375fe2a01/config/locales/ru/categories.yml

1000miles:
@schultyy Thanks, we only touched the reference file, all others will be updated automatically in a follow-up process when pulling the new translations.

1000miles:
Closed because this task is also done in PR #312 with the last translations import.

https://github.com/sozialhelden/wheelmap/blob/57da039aa35fe551ac52f6ac195171011c32ea4e/config/locales/de/categories.yml#L73

PR #307, opened 2016-06-17
Implement eslint and fix code accordingly.

lennerd:
This PR implement eslint including configuration and adding some dev dependencies.

The changeset looks huge, but it's actually just fixing some linting errors/JS code style issues.

* [x] Fix eslint errors
* [x] Fix tests
* [x] Update jest test framework to work with es6
* [x] Use es6 everywhere


app/assets/javascripts/NewNodeWidget/Breadcrumbs.js
```diff
@@ -1,16 +1,14 @@
...
-const classNames = require('classnames');
+import React, { Component } from 'react';
+import classNames from 'classnames';
 
-const I18n = require('../common/I18n');
-const Section = require('./models/Section').default;
-const { OVERVIEW } = require('./models/sections');
+import I18n from '../common/I18n';
+import Section from './models/Section';
+import { OVERVIEW } from './models/sections';
```
schultyy: Is `OVERVIEW` a constant?

app/assets/javascripts/NewNodeWidget/Section/Accessibility.js
```diff
@@ -1,19 +1,18 @@
-const React = require('react');
-const { bindActionCreators } = require('redux');
-const { createStructuredSelector } = require('reselect');
-const { connect } = require('react-redux');
+import React, { PropTypes } from 'react';
```
schultyy: Shouldn't this be `import { React, PropTypes } from 'react';`?

app/assets/javascripts/NewNodeWidget/Section/Address.js
```diff
@@ -1,27 +1,27 @@
...
-const { nodeTypesSelector } = require('../../common/selectors/nodeTypes');
-const NodeType = require('../../common/models/NodeType');
-const Node = require('../../common/models/Node');
-
-const { func, instanceOf, bool, listOf, number, string } = React.PropTypes;
-const { latlng } = Leaflet.PropTypes;
-const { immutableMapOf, immutableListOf } = require('../../common/types');
-
-class AddressSection extends React.Component {
+import React, { Component, PropTypes } from 'react';
```
schultyy: Shouldn't this be `import { React, Component, PropTypes } from 'react';`?

app/assets/javascripts/NewNodeWidget/Breadcrumbs.js
lennerd: Yes, there is a fixed set of sections which are kind of used as constants. Thats why I use upper case for their names. Do you have a better way in mind?

app/assets/javascripts/NewNodeWidget/Section/NameCategory.js
```diff
@@ -36,19 +46,23 @@ class NameCategorySection extends React.Component {
         <div>{/* Needed for having only one section */}
           <Form.ControlGroup labelFor="name" labelScope="activerecord.attributes.poi.name">
             <Form.Controls>
-              <Form.Input name="name"
-                          onChange={this.onChangeName.bind(this, node)}
-                          value={node.name}
-                          errors={errors.get('name')}/>
+              <Form.Input
+                name="name"
+                onChange={this.onChangeName}
```
schultyy: is `bind(this, node)` not necessary anymore?

app/assets/javascripts/NewNodeWidget/Section/NameCategory.js
```diff
@@ -36,19 +46,23 @@ class NameCategorySection extends React.Component {
...
             <Form.Controls>
-              <NodeTypeSelect categories={categories}
-                              nodeTypes={nodeTypes}
-                              onChange={this.onChangeNodeType.bind(this, node)}
-                              value={node.nodeType}
-                              errors={errors.get('nodeType')}/>
+              <NodeTypeSelect
+                categories={categories}
+                nodeTypes={nodeTypes}
+                onChange={this.onChangeNodeType}
```
schultyy: Is `.bind(this, node)` not necessary anymore?

app/assets/javascripts/NewNodeWidget/Section/Accessibility.js
lennerd: React is exported as default and used when the JSX code is compiled. We can also just use `import React from 'react';` and in code `const { bool } = React.PropTypes;`. 

In terms how the React library is structured, `import { React, PropTypes } from 'react';` would mean, we can use `React.React`, but this is not the case.

app/assets/javascripts/NewNodeWidget/Section/Address.js
lennerd: See above.

app/assets/javascripts/NewNodeWidget/Section/SimilarNodes.js
```diff
@@ -1,47 +1,50 @@
...
-               onClickAction={onClickAction}
-               loading={loading}>
-        <ul className="nodes-new-content-section--similar-list">
-          {items}
-        </ul>
-      </Section>
+function SimilarNodesSection({ similarNodes, loading, onClickAction }) {
+  const items = [];
+
+  similarNodes.forEach(node => {
```
schultyy: you could refactor the `similarNodes.forEach(node => { ... })` also into an `const items = similarNodes.map(node => { ... })`.

app/assets/javascripts/NewNodeWidget/Section/NameCategory.js
lennerd: No, the trick is to use an arrow function inside the class definition:

``` js
onChangeName = (event) => {
  const { node, onChange } = this.props;

  onChange(node.set('name', event.target.value));
};
```

This way the context is passed through and no `bind` is needed. This solution is much faster and in my opinion also much cleaner and clearer.

This gets transpiled to a function defined in the constructor of the class in < ES5.

app/assets/javascripts/NewNodeWidget/Section/NameCategory.js
lennerd: See above.

app/assets/javascripts/NewNodeWidget/Section/SimilarNodes.js
lennerd: Unfortunately not. The `similarNodes` object is a [immutable map from Immutable.js](https://facebook.github.io/immutable-js/docs/#/Map) and therefore not a native array.

We could of course write something like:
``` js
const items = similarNodes.toArray().map((node) => { /*...*/ });
```

Not sure if this is better ...

lennerd: You got me thinking. I did a little bit of research and found out, that Immutable data structures can be used as children. 🎉 😁 Will fix this.

app/assets/javascripts/NewNodeWidget/sagas/activateSection.js
```diff
@@ -3,21 +3,21 @@ import find from 'mout/collection/find';
...
 
 export default function*() {
   while (true) {
-    const { payload: { section: sectionId } } = yield take(ENTER_CONTENT),
-      section = find(sections, section => section.id === sectionId);
+    const { payload: { section: sectionId } } = yield take(ENTER_CONTENT);
+    const section = find(sections, section => section.id === sectionId);
 
     // Redirect to name and category (first) section if given section is invalid.
     if (section == null) {
```
schultyy: here you can use `===` instead of `==`

app/assets/javascripts/NewNodeWidget/sagas/updateMap.js
```diff
@@ -2,43 +2,35 @@ import { take, put, cancel, fork, call } from 'redux-saga/effects';
...
     yield call(delay, 300);
 
-    const address = yield call([node, node.address]),
-      feature = yield call(photon.geocode, address);
+    const address = yield call([ node, node.address ]);
+    const feature = yield call(geocode, address);
 
     // Restart daemon if no feature was found.
-    if (feature == null)
+    if (feature == null) {
```
schultyy: here you can use `===` instead of `==`

app/assets/javascripts/NewNodeWidget/sagas/updateMap.js
```diff
@@ -55,15 +47,27 @@ export function *debounceUpdateMap() {
       const { payload: node } = yield take(CHANGE_NODE_ADDRESS);
 
       // Cancel old update map task (debounce).
-      if (updateMapTask != null)
+      if (updateMapTask != null) {
```
schultyy: here you can use `!==` instead of `!=`

app/assets/javascripts/WidgetBuilder/App.js
```diff
@@ -1,26 +1,31 @@
...
+import createStore from './helpers/createStore';
+import WidgetBuilder from './WidgetBuilder';
+import Widget from './models/Widget';
+import widgetBuilderReducer from './reducer';
+import { init } from './actions';
+
+const { object } = PropTypes;
+
+function WidgetBuilderApp({ widget }) {
+  widget = new Widget(camelize(widget));
```
schultyy: This one is a bit strange. The function gets a `widget` and it is overwritten immediately with a new widget falling out of `Widget`. I would suggest to use a different variable name here to make the difference clear.

app/assets/javascripts/NewNodeWidget/sagas/activateSection.js
lennerd: This way `section == null` is true both when section is undefined or null. But maybe its easier to just check if `section` is simply truthy?

app/assets/javascripts/NewNodeWidget/sagas/updateMap.js
lennerd: See above.

app/assets/javascripts/NewNodeWidget/sagas/updateMap.js
lennerd: See above.

app/assets/javascripts/WidgetBuilder/App.js
lennerd: True. Will think about it.

lennerd:
@schultyy Thanks for the very extensive feedback!!! 👍 😁

schultyy:
I'm not done yet :)

app/assets/javascripts/common/Photon.js
```diff
@@ -27,51 +24,67 @@ class Photon extends React.Component {
...
+    const { lang, limit, url } = this.props;
+
     // Abort old feature request
-    if (this.featureRequest != null)
+    if (this.featureRequest != null) {
       this.featureRequest.abort();
+    }
 
+    // Remove jQuery
     this.featureRequest = $.ajax({
```
schultyy: If jQuery is used only to make the XHR here, maybe it makes sense to convert this piece of code to use the new [`fetch`](https://github.com/github/fetch) API?

app/assets/javascripts/NewNodeWidget/Breadcrumbs.js
schultyy: No, if they're constants uppercased names are fine.

app/assets/javascripts/NewNodeWidget/Section/Accessibility.js
schultyy: I see.

app/assets/javascripts/NewNodeWidget/Section/NameCategory.js
schultyy: Nice trick :)

app/assets/javascripts/NewNodeWidget/sagas/activateSection.js
schultyy: Hmm, another case: Could it be that `section` could be undefined?

app/assets/javascripts/common/Photon.js
lennerd: Maybe it makes sense to simple use Vanilla JS here, yes. Unfortunately the `fetch` API has no abort method which is essential to stop old request for old search inputs.

Will fix this.

app/assets/javascripts/NewNodeWidget/sagas/activateSection.js
lennerd: You're true. In this case, as `find` returns a section or `undefined`, we can check against undefined.

lennerd:
@schultyy Thanks again for your feedback. Added some changes and even a few more eslint rules.

lennerd:
Rebased and fixed two errors with setParams and error layout in new node form.

Issue #296, opened 2016-06-13
Upgrade to latest Rails 4.2.x release


schultyy:
In order to upgrade the application to the latest 4.2.x release, the following things need to be taken into consideration:

Rails 4.1:

- use `secrets.yml`
- CSRF protection from remote `<script>` tags

Rails 4.2: 
- add `responders` gem to keep `respond_to` functionality
- Check `*_filter` methods, they've been renamed to `*_action`

These points above are not complete. I browsed through http://guides.rubyonrails.org/4_1_release_notes.html and http://guides.rubyonrails.org/4_2_release_notes.html#upgrading-to-rails-4-2
and checked if points listed do concern us.

schultyy:
For more details we also need to consult http://guides.rubyonrails.org/upgrading_ruby_on_rails.html

schultyy:
For Rails 4.2 we need to update some Gems. `bundle update rails` with Rails 4.2 in the `Gemfile` produces this:
```
[0] % bundle update rails
Fetching gem metadata from https://rubygems.org/
Fetching version metadata from https://rubygems.org/
Fetching dependency metadata from https://rubygems.org/
Resolving dependencies....................................................
Bundler could not find compatible versions for gem "actionpack":
  In Gemfile:
    actionpack-page_caching (>= 1.0.2, ~> 1.0) was resolved to 1.0.2, which depends on
      actionpack (< 5, >= 4.0.0)

    formtastic was resolved to 3.1.3, which depends on
      actionpack (>= 3.2.13)

    inherited_resources (~> 1.6.0) was resolved to 1.6.0, which depends on
      actionpack (< 5, >= 3.2)

    inherited_resources (~> 1.6.0) was resolved to 1.6.0, which depends on
      actionpack (< 5, >= 3.2)

    kaminari (~> 0.14) was resolved to 0.16.3, which depends on
      actionpack (>= 3.0.0)

    rails (~> 4.2.0) was resolved to 4.2.0, which depends on
      actionpack (= 4.2.0)

    rails (~> 4.2.0) was resolved to 4.2.0, which depends on
      actionpack (= 4.2.0)

    activeadmin (~> 1.0.0.pre2) was resolved to 1.0.0.pre2, which depends on
      ransack (~> 1.3) was resolved to 1.7.0, which depends on
        actionpack (>= 3.0)

    routing-filter was resolved to 0.4.0.1, which depends on
      actionpack (< 4.2)

[...]

PR #308, opened 2016-06-20
Update to Rails 4.1

schultyy:
PR for #296 

This updates the wheelmap application to Rails 4.1. Most notable change: 
All secrets now go into `config/secrets.yml`. I moved all contents from `config/application.yml` over into the new file. Therefore `config/application.yml` is obsolete and deleted.
If `application.yml` is already in place the application still works as expected but newly added secrets do go into `secrets.yml` now.

/cc @lennerd @1000miles


justahero:
LGTM

1000miles:
Thanks for the PR! :)

PR #312, opened 2016-06-21
Import new translation files

1000miles:
This PR imports all new translations that have been added on transifex. We had to force pull the translation files via `tx pull -f` because somehow transifex seemed to always skip a couple of files even though we could see the updated files on transifex.

With the newly imported files this fixes the `community_center` error in issue #282. And it also makes the PR #306 redundant, therefore PR #306 will be closed without merge.


schultyy:
LGTM

PR #320, opened 2016-06-30
Remove x-frame-options header from iframe embeds response

1000miles:
This PR fixes the broken embedding of iframes by removing the x-frame-options header from http-responses.


schultyy:
It provides a fix for the case when a user wants to use `wheelmap.org/embed` in an iFrame. If they want to embed `wheelmap.org` in an iFrame, this will not work, which is on purpose to prevent [Clickjacking](https://en.wikipedia.org/wiki/Clickjacking)

justahero:
Looks fine to me.

1000miles:
Thanks, @schultyy & @justahero ! :) Will deploy it on monday.

Issue #90, opened 2016-01-11
Update dev-env for MacOS, Ubuntu & Vagrant in README.md


1000miles:
It takes a lot of time to research, configure and (re-)setup the project wheelmap dependent on the different environments (e.g. Ubuntu, Vagrant, etc.) and packages that are used. 

Currently the [README file](https://github.com/sozialhelden/wheelmap/blob/master/README.md) contains installation instructions mainly for Homebrew which is rather relevant for Mac Users, another example are the instructions for rbenv which does not apply for rvm users or users with no Ruby version managers. An instruction update for different environments might help new contributors to get started faster with this project.

We will write/edit install instructions for:
- [x] MacOS
- [x] Ubuntu
- [x] Vagrant

Hoverbear:
https://github.com/sozialhelden/wheelmap/pull/431 contributes both a working Vagrant box as well as some brief instructions.

1000miles:
PR #431 merged to master today. Closing.

PR #321, opened 2016-07-01
Add & edit install instructions for MacOS & Ubuntu

1000miles:
This PR does following:
- adds install instructions for Ubuntu
- improves install instructions for MacOS
- solves partly issue: #90 

We removed the rbenv installation part and will move it to the wiki page. In a later phase we will also add infos about how to set up the project with vagrant.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,144 +7,215 @@ Wheelmap.org is an online map to search, find and mark wheelchair-accessible pla
 
 This Repository is the source code of the website [Wheelmap.org](http://wheelmap.org/).
 
-## Installation
+## I. Installation
+
 
 ### Requirements
 
-If you are working on a Mac, please install [Homebrew](http://brew.sh/).
+- Ruby 2.2.2
+- Bundler
+- MySQL
+- ImageMagick
+- PhantomJS
+- node / npm
 
-Then install the following required tools:
 
-#### `git`, `wget`
+Before you start, please make sure you have [Homebrew](http://brew.sh/) for Mac or [aptitude](http://packages.ubuntu.com/search?keywords=aptitude) for Ubuntu installed.
 
-    brew install git wget
+#### Install essential software packages:
 
-#### Latest Ruby 2.2.2 via rbenv
+MacOS
 
-    brew install rbenv ruby-build
-    rbenv install 2.2.2
-    echo 'export PATH="/usr/local/bin:$PATH"' >> ~/.bash_profile
-    echo 'eval "$(rbenv init -)"' >> ~/.bash_profile
+```
+$ brew install git wget
+```
 
-Restart your shell and install bundler:
+Ubuntu 
 
-    rbenv global 2.2.2
-    gem install bundler
-    rbenv rehash
+```
+$ sudo apt-get update
+$ sudo apt-get install -y git-core wget 
 
-### Dependencies
+# more libraries
+$ sudo apt-get install -y curl zlib1g-dev build-essential libssl-dev libreadline-dev libyaml-dev libxml2-dev libxslt1-dev libcurl4-openssl-dev python-software-properties libffi-dev
+```
 
-#### MySQL
+#### Install Ruby:
 
-    brew install mysql
-    mkdir -p ~/Library/LaunchAgents
-    ln -sfv /usr/local/opt/mysql/*.plist ~/Library/LaunchAgents
-    launchctl load ~/Library/LaunchAgents/homebrew.mxcl.mysql.plist
+MacOS / Ubuntu
 
-#### ImageMagick
+First, install the Ruby Package Manager of your choice:
 
-    brew install imagemagick
+- [rbenv](https://github.com/rbenv/rbenv)
+- [rvm](https://rvm.io/rvm/install)
 
-#### PhantomJS
+Dependent on your choice, please install Ruby with either:
+
+```
+$ rbenv global 2.2.2
+$ rbenv rehash
+```
+
+or:
+
+```
+$ rvm install 2.2.2
+$ rvm use 2.2.2 
+```
+
+#### Install Bundler: 
+
+MacOS / Ubuntu
+
+```
+$ gem install bundler
+```
+
+#### Install MySQL:
+
+MacOS
+
+```
+$ brew install mysql
+$ mkdir -p ~/Library/LaunchAgents
+$ ln -sfv /usr/local/opt/mysql/*.plist ~/Library/LaunchAgents
+$ launchctl load ~/Library/LaunchAgents/homebrew.mxcl.mysql.plist
+```
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -7,144 +7,219 @@ Wheelmap.org is an online map to search, find and mark wheelchair-accessible pla
...
+```
+$ brew install git wget
+```
 
-#### Latest Ruby 2.2.2 via rbenv
+Ubuntu 
+
+```
+$ sudo apt-get update
+$ sudo apt-get install -y git-core wget
```
schultyy: Is there any reason we depend on `git-core`? If not can you change `git-core` to `git`?

README.md
```diff
@@ -7,144 +7,219 @@ Wheelmap.org is an online map to search, find and mark wheelchair-accessible pla
...
 
-#### Latest Ruby 2.2.2 via rbenv
+Ubuntu 
+
+```
+$ sudo apt-get update
+$ sudo apt-get install -y git-core wget 
+
+# more libraries
+$ sudo apt-get install -y curl zlib1g-dev build-essential libssl-dev libreadline-dev libyaml-dev libxml2-dev libxslt1-dev libcurl4-openssl-dev python-software-properties libffi-dev
```
schultyy: I would suggest to install all dependencies in one step. There's no need to have two install steps here.

README.md
```diff
@@ -7,144 +7,219 @@ Wheelmap.org is an online map to search, find and mark wheelchair-accessible pla
...
+
+#### Install MySQL:
+
+MacOS
+
+```
+$ brew install mysql
+$ mkdir -p ~/Library/LaunchAgents
+$ ln -sfv /usr/local/opt/mysql/*.plist ~/Library/LaunchAgents
+$ launchctl load ~/Library/LaunchAgents/homebrew.mxcl.mysql.plist
```
schultyy: I'm not sure about these lines:
```
$ mkdir -p ~/Library/LaunchAgents
$ ln -sfv /usr/local/opt/mysql/*.plist ~/Library/LaunchAgents
$ launchctl load ~/Library/LaunchAgents/homebrew.mxcl.mysql.plist
```
I would vote to remove them because they set up MySQL to run as a daemon permanently even if it's not used.

README.md
```diff
@@ -7,144 +7,219 @@ Wheelmap.org is an online map to search, find and mark wheelchair-accessible pla
...
+
+```
+$ wget http://download.geofabrik.de/europe/germany/berlin-latest.osm.bz2
+$ bzcat berlin-latest.osm.bz2 | bundle exec rake osm:import
+```
 
-    ```
-    bundle exec rake db:seed
-    ```
+##### Install node and all node javascript dependencies:
```
schultyy: Install all `JavaScript` dependencies

README.md
```diff
@@ -7,144 +7,219 @@ Wheelmap.org is an online map to search, find and mark wheelchair-accessible pla
...
-9. And get some POI data into the database:
+MacOS 
 
-    ```
-    wget http://download.geofabrik.de/europe/germany/berlin-latest.osm.bz2
-    bzcat berlin-latest.osm.bz2 | bundle exec rake osm:import
-    ```
+```
+$ brew install node
+$ npm install
```
schultyy: Depending on the Node version `npm` might be outdated. Can you add the following line before `npm install`?

```
$ npm install -g npm
```

This installs the latest version of npm.

README.md
```diff
@@ -7,144 +7,219 @@ Wheelmap.org is an online map to search, find and mark wheelchair-accessible pla
...
 
-10. Install all node javascript dependencies:
+Ubuntu 
 
-    ```
-    npm install
-    ```
+```
+$ sudo apt-get install -y nodejs
+$ sudo apt-get install -y npm
```
schultyy: Same as above: Can you add the line about update npm here as well?

README.md
```diff
@@ -7,144 +7,219 @@ Wheelmap.org is an online map to search, find and mark wheelchair-accessible pla
...
 
 ### Requirements
 
-If you are working on a Mac, please install [Homebrew](http://brew.sh/).
+- Ruby 2.2.2
+- Bundler
+- MySQL
+- ImageMagick
+- PhantomJS
+- node / npm
```
schultyy: We need to be concrete here which Node version is required. @lennerd can you tell us which version you use right now?

lennerd: As we use [Browserify](https://github.com/substack/node-browserify) to bundle all resources it should be no problem to use the latest Node version available.

I ran my setup with Node v5.11.0, but older versions should work fine, too. Travis is using v4.4.7. Does this help.

1000miles: Thanks, @lennerd! So should we say `node >= 4.4.7` to be on the save side because of Travis or should we leave out the version number?

lennerd: @schultyy Do you have any experience about Node JS backwards compatibility? I have the feeling Node is very safe in terms of version switch, right? Maybe we can leave out the number for now till we come across a problem?

schultyy: Depends. I wouldn't recommend to use anything < 4 because they're out of support soon-ish. As far as I remember Node 6.x introduced breaking changes so that many modules didn't work anymore. Especially those with native extensions. What about recommending `>= 4.x & < 6`?

lennerd: I think >= 4.x is enough as we will not work with native extensions when using browserify (because they do not exist in the browser). If we anyone comes across a problem we can always adjust the number. What do you think?

schultyy: Sure in the browser there are no native extensions. But it could be that one of the modules used by browserify has native extensions.

schultyy: We can always adjust the number. I just saw that we also have the Node version specified in `package.json`: https://github.com/sozialhelden/wheelmap/blob/master/package.json#L49. If we change it we should change it there as well.

lennerd: Good catch. So let's stick with Node >= 4.1.

1000miles: Alright, thank you for your feedback, @schultyy & @lennerd.

README.md
1000miles: Thanks, I will move this part to the wiki for optional MqSQL configs.

README.md
```diff
@@ -187,20 +181,22 @@ $ wget http://download.geofabrik.de/europe/germany/berlin-latest.osm.bz2
 $ bzcat berlin-latest.osm.bz2 | bundle exec rake osm:import
 ```
 
-##### Install node and all node javascript dependencies:
+##### Install all `Javascript` dependencies:
```
schultyy: No need for backticks here.

schultyy: And it's JavaScript :)

schultyy:
LGTM 🙆

lennerd:
Great updates and a very handy README! 🙇

PR #322, opened 2016-07-05
Fix/new node feedback

lennerd:
Fixes #195.

PR #323, opened 2016-07-05
Use `pretty_print` for JS translations assets.

lennerd:
This PR does not change anything else then the `pretty_print` option.

#322 needs to merged first as this PR is based on it.


schultyy:
lgtm

PR #343, opened 2016-07-25
Add map zoom alert.

### LINKED ISSUE #316: Add notice "Zoom in closer to see all places"
Svenja suggests:

As a user who is looking for a place by zooming into the map, I am not aware that certain places appear only when I zoom in closest to the highest zoom level.

I would find it helpful to get informed of this particularity in a info bubble saying "Zoom in closer to see all places" that stays visible until the highest zoomlevel is displayed.

lennerd:
Fixes #316.

screenshot 2016-07-25 14 11 48

screenshot 2016-07-25 14 12 04

When clicking on the message, the user get zoomed in to the maximum zoom level available. The message disappears as soon as he reaches this zoom level.

* [x] Fix position.
* [x] Use translations.

### DIFF
--- a/app/assets/javascripts/i18n/ar.js
+++ b/app/assets/javascripts/i18n/ar.js
@@ -839,7 +839,8 @@ I18n.translations["ar"] = I18n.extend((I18n.translations["ar"] || {}), {
         },
         "help": "هل وصول الكرسي المتحرك سهل؟ (بادر بتقديم المساعدة)",
         "more": "المزيد..."
-      }
+      },
+      "zoom_alert": "Zoom in closer to see more places."
     }
   },
   "how?": "كيف ؟",
--- a/app/assets/javascripts/i18n/bg.js
+++ b/app/assets/javascripts/i18n/bg.js
@@ -839,7 +839,8 @@ I18n.translations["bg"] = I18n.extend((I18n.translations["bg"] || {}), {
         },
         "help": "Достъпно за инвалидни колички? (Помощ)",
         "more": "повече..."
-      }
+      },
+      "zoom_alert": "Zoom in closer to see more places."
     }
   },
   "how?": "Как?",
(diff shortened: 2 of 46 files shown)


lennerd:
An alternative to zoom in to the highest zoom level would be simple step by step zoom, which does need more clicks but is maybe less confusing. @schultyy Is deploying this on Staging right now. @holgerd What do you think?

holgerd:
@lennerd I like the idea of multiple clicks. 
How about putting the text next to the zoom buttons on the left side? Would maybe make the context of this message clearer.

lennerd:
@holgerd What do you think about changing the text to "Zoom in closer to see more places." to fit to the proposed interaction?

lennerd:
@holgerd The full zoom is on Staging now.

lennerd:
@holgerd Changes are deployed on Staging. Please review.

screenshot 2016-07-25 15 03 05

holgerd:
LGTM
I like.

schultyy:
I'm not sure if it makes sense to have the english translation in all translation files. If a translation is missing it should default to the default translation which should be either German or English, right?

lennerd:
@schultyy and I further discussed this. The provided translations files are automatically generated by the `i18n-js` gem. The source translation files do not contain mixed translations. And as we use Transifex, we don't need Rails to show unavailable translations.

@Svenyo FYI: We added a new translation "Zoom in closer to see more places."

holgerd:
related #344

Svenyo:
ok, the english wording is fine. However, I couldn't find the new string on transifex.

lennerd:
Sorry, pushed from the wrong Branch. The translation is now on Transifex: https://www.transifex.com/sozialhelden/wheelmap-website/translate/#de/relaunchyml/93865875?translated=no

Svenyo:
ok

PR #350, opened 2016-08-01
Use English as default locale.

### LINKED ISSUE #344: If a translation is missing it should default to the English translation
If a translation is missing it should default to the default translation which should be either German or English

This came up in 
https://github.com/sozialhelden/wheelmap/pull/343#issuecomment-234968989

lennerd:
Fixes #344.

This PR changes the default locale to English (previously German).

The only problem I can see is a user visting the page without a cookie and a previously stored locale. When he visits wheelmap.org/map he will probably see the page in English. I'm not quite sure how solid the switch on the locale HTTP header is though: https://github.com/sozialhelden/wheelmap/blob/02412ab4465790ed6ab1c63171561e0013c6aab9/lib/rack_i18n_locale_switcher.rb#L68-L79

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -28,7 +28,7 @@ module Wheelmap
     config.middleware.use(Rack::I18nLocaleSwitcher)
     config.i18n.fallbacks = true
     config.i18n.load_path += Dir[Rails.root.join('config', 'locales', '**', '*.{rb,yml}')]
-    config.i18n.default_locale = :de
+    config.i18n.default_locale = :en
     config.i18n.enforce_available_locales = false
 
     config.i18n.available_locales = [:de]


Hoverbear:
Looks like many of these test failures are due to expecting German strings, example: https://github.com/sozialhelden/wheelmap/blob/40c37c48731d8b04f3b3c43848f02f3563b83796/spec/features/signup_spec.rb#L27

holgerd:
What effects will have this on our URL structure (from a SEO perspective)? 

Today I see live:

- https://wheelmap.org/en/nodes/1296421604 --> 200 OK
- https://wheelmap.org/de/nodes/1296421604 --> 302 REDIRECT to https://wheelmap.org/nodes/1296421604

How will this change?

## Option 1: always redirect to language-specific URL

- https://wheelmap.org/nodes/1296421604 --> 301 REDIRECT to https://wheelmap.org/en/nodes/1296421604
-  https://wheelmap.org/de/nodes/1296421604 --> 200 OK

## Option 2: Have English as the non-locale URL

- https://wheelmap.org/en/nodes/1296421604 --> 301 REDIRECT to https://wheelmap.org/nodes/1296421604
- https://wheelmap.org/de/nodes/1296421604 --> 200 OK

I suggest Option 1.

@Xylakant @lennerd Am I missing something here?

holgerd:
We will not do this as discussed in https://github.com/sozialhelden/wheelmap/issues/344#issuecomment-253453913

Issue #351, opened 2016-08-05
New logging strategy


schultyy:
When the wheelmap application is going to be deployed on the new servers the logging strategy should be changed.
Right now the application logs to `log/production.log` or `log/staging.log` depending on the environment. We would like to change this to the following:

The application logs only to stdout and stderr respectively. Then via systemd we're able to capture the output and store it in the journal.

Hoverbear:
This was addressed in #353 however we're waiting backend changes.

Xylakant:
done and merged to the main infra-rebuild branch

PR #353, opened 2016-08-05
Log to stdout

Hoverbear:
Regarding #351. I wasn't sure if the rack loggers should be directed as well, so please review.


schultyy:
Thank you for the PR 🎆 
But we need to wait with merging until we have the new servers up and running :)

Issue #354, opened 2016-08-08
MySQL 5.7: rake db:migrate fails


schultyy:
When we upgrade to MySQL 5.7 we do have some problems with the database:

When running `rake db:migrate`:
```
vagrant@vagrant:~/wheelmap$ rake db:migrate
== 20110107131649 InitialSchema: migrating ====================================
-- create_table("active_admin_comments")
   -> 0.0110s
-- add_index("active_admin_comments", ["author_type", "author_id"], {:name=>"index_active_admin_comments_on_author_type_and_author_id"})
   -> 0.0108s
-- add_index("active_admin_comments", ["namespace"], {:name=>"index_active_admin_comments_on_namespace"})
   -> 0.0085s
-- add_index("active_admin_comments", ["resource_type", "resource_id"], {:name=>"index_admin_notes_on_resource_type_and_resource_id"})
   -> 0.0095s
-- create_table("admin_users")
   -> 0.0106s
-- add_index("admin_users", ["confirmation_token"], {:name=>"index_admin_users_on_confirmation_token", :unique=>true})
   -> 0.0131s
-- add_index("admin_users", ["email"], {:name=>"index_admin_users_on_email", :unique=>true})
   -> 0.0119s
-- add_index("admin_users", ["reset_password_token"], {:name=>"index_admin_users_on_reset_password_token", :unique=>true})
   -> 0.0155s
-- create_table("categories")
   -> 0.0177s
-- create_table("counters")
   -> 0.0100s
-- create_table("delayed_jobs")
   -> 0.0143s
-- add_index("delayed_jobs", ["priority", "run_at"], {:name=>"delayed_jobs_priority"})
   -> 0.0104s
-- create_table("experiments")
   -> 0.0116s
-- create_table("node_types")
   -> 0.0115s
-- add_index("node_types", ["id", "category_id"], {:name=>"index_node_types_on_id_and_category_id"})
   -> 0.0138s
-- add_index("node_types", ["osm_key", "osm_value"], {:name=>"index_node_types_on_osm_key_and_osm_value"})
   -> 0.0123s
-- create_table("photos")
   -> 0.1217s
-- create_table("pois", {:id=>false, :options=>"ENGINE=MyISAM DEFAULT CHARSET=utf8 COLLATE=utf8_unicode_ci"})
   -> 0.0026s
-- add_index("pois", ["geom"], {:name=>"index_pois_on_geom", :spatial=>true})
   -> 0.0035s
[...]

PR #358, opened 2016-08-09
Multipolygon -> Geometry for MySQL 5.7

### LINKED ISSUE #355: MySQL 5.7: Failing tests with Cannot get geometry object from data you send to the GEOMETRY field
When running the tests with MySQL 5.7, we encounter the following error:

```
Mysql2::Error: Cannot get geometry object from data you send to the GEOMETRY field: INSERT INTO `regions` (`created_at`, `grenze`, `lft`, `name`, `rgt`, `slug`, `updated_at`) VALUES ('2016-08-08 14:04:51', GeomFromWKB(0x00000000030000000100000005402a2d36f7e3d1cc404a2b463497b741402b8595feda6613404a2b463497b741402b8595feda6613404a567620ee8d11402a2d36f7e3d1cc404a567620ee8d11402a2d36f7e3d1cc404a2b463497b741,0), 1, 'Berlin', 2, 'berlin', '2016-08-08 14:04:51')
```

Hoverbear:
Might fix #355 #354 .

Please review and test. I tested in an Ubuntu docker container and encountered no issues, but I'm also not entirely sure what might be affected by this change.

### DIFF
--- a/db/structure.sql
+++ b/db/structure.sql
@@ -238,7 +238,7 @@ CREATE TABLE `queued_nodes` (
 CREATE TABLE `regions` (
   `id` bigint(20) NOT NULL AUTO_INCREMENT,
   `name` varchar(255) COLLATE utf8_unicode_ci DEFAULT NULL,
-  `grenze` multipolygon NOT NULL,
+  `grenze` geometry NOT NULL,
   `created_at` datetime DEFAULT NULL,
   `updated_at` datetime DEFAULT NULL,
   `parent_id` bigint(20) DEFAULT NULL,


Xylakant:
hmm. this is problematic. We'd need a migration to change the existing database and a migration path for the existing data. What's the difference between Multipolygon and Geometry?

Xylakant:
the mysqlspatial adapter doesn't seem to support multipolygon.

Xylakant:
it looks like the the https://github.com/rgeo/activerecord-mysql2spatial-adapter is not well supported since at least a year.

Hoverbear:
I kind of just "discovered" this while looking around. This is definitely not a complete pull request if we choose to go this route. :)

Hoverbear:
Hmm... Is there a better alternative?

Hoverbear:
Closing due to staleness.

Issue #181, opened 2016-04-11
On after signup edit page: alert message duplicate appears when providing email address


1000miles:
**Case:**

- I have registered as a new user at openstreetmap.org
- I am logged into my new account
- I am on after signup edit page (/profile/after_signup_edit)
- I have type in my email address
- I click on Fertigstellen button

**Then I get this alert message:**

<img width="1271" alt="00000072" src="https://cloud.githubusercontent.com/assets/4077916/14432412/591bf36e-000a-11e6-9faa-c39f58147ca4.png">

**When I click on OK, a second alert pops up with a similiar message:**

<img width="1277" alt="00000074" src="https://cloud.githubusercontent.com/assets/4077916/14432439/787f2cda-000a-11e6-9d14-040d401b4987.png">

lennerd:
Cannot reproduce in Chrome and Firefox. Instead of seeing a second alert I get redirected to the map. @1000miles Does this issue still appear in your setup?

lennerd:
I will remove this alert as it simply shows the content also shown in a flash message on map after the redirect.

Issue #179, opened 2016-04-08
Needs Revision: Signup, Login, Password Recovery & Profile Editing


1000miles:
The signup, login, password recovery and profile editing process needs a revision because at some parts the redirections, flash messages and email messages don't follow a logical concept anymore since the signup/login authentication has been moved to openstreetmap.

This ticket is a parent task to collect all the subtasks. 

**Example:**

![new_email_confirmation_page](https://cloud.githubusercontent.com/assets/4077916/14377580/b4b5a004-fd70-11e5-91ad-5d1bfed29eff.png)

- [ ]  On profile edit page: Flash message[success] not working properly #176 
- [ ]  On after signup edit page: form validation not working properly #175

holgerd:
related: #349

holgerd:
on hold, waiting for PR #359 to review this.

Issue #175, opened 2016-04-04
On after signup edit page: form validation not working properly


1000miles:
**Case:**
* I am logged in 
* I am on after signup edit page (`/profile/after_signup_edit`)
* I leave the email input field empty (even though it is marked as required) or I type in a non-valid email address with `dfdgfdgdg`
* I click on `Fertigstellen` button

=> I should get an error message which tells me to provide an email address first in order to complete the signup process, instead I get the success message: `Du bekommst in wenigen Minuten eine E-Mail mit einem Link, um deine E-Mail-Adresse zu bestätigen.`

**In the browser inspector I can find this:**
`<div id="user_email_input" class="email input required stringish form-group">`

@lennerd It seems the form validation doesn't work properly? 

<img width="1275" alt="after_signup_edit_page_message" src="https://cloud.githubusercontent.com/assets/4077916/14241838/24fb682e-fa4e-11e5-88fe-db3160c5db4f.png">

lennerd:
Yeah, that seems to be a bug. I will take a look at this as soon as #105 is deployed.

1000miles:
:+1: Thank you for looking into it.

lennerd:
Do you created a new account or did you login with an existing one?

holgerd:
on hold, waiting for PR #359 to review.

PR #359, opened 2016-08-15
Fix/login signup

### LINKED ISSUE #176: On profile edit page: Flash message[success] not working properly
**Given:**
- I have registered at openstreetmap.org
- I am logged into my account
- I am on my profile page (/profile)

**Steps on profile edit page:**
- I type in a new password
- I repeat my new password
- I type in my email address (which is empty by default) 
- I click on Speichern

=> I should see a flash message[success] in order to be informed, if my profile update was successful, instead I don't get any hints.

**In the browser inspector I can see that the flash message is implemented, but set to non-visible**
`<div id="success_div_id" class="notification success" style="display:none"></div>`

@lennerd Do you have an idea why it is set to non-visible here?

profile_edit_page_flash_success

**As a comparison, the flash message[error] seems to work properly when not proving an email address here:**

profile_edit_page_flash_error

**Browser inspector:**
`<div id="alert_div_id" class="notification alert" style=""></div><span>Email muss ausgefüllt werden, wenn ein Passwort gesetzt wird</span></div>`

lennerd:
This PR tries to fix #176, #175, #179 and #181.

**What did change?**

* The user email address is mandatory now. If a user is logged in for the first time or did not set an email address yet, we **always** show him `after_signup_update` page to add a email address.<br />*This part is really critical I think. We need to be sure to force user to have a email address. There is a quite big opportunity users get angry or frustrated because they are forced to give data to us.*<br />*We also need to check if this has any implications for the apps on the mobile devices.*
* The user password was only needed so users were able to login via their mobile devices. As we refactored this process completely via an integrated WebView, I removed the password field completely.
* The `after_signup_update` view cannot be skipped anymore.
* I remove some of the duplicate flash messages and JavaScript alerts, which were shown when the user clicked on the submit button or reached the front page.

**What needs to be done?**

* [ ] Fix tests.

**Note:**

* I thought a while about making `after_signup_update` "beautiful again" and I decided to don't touch it. The problem is that we do not have any kind of Style Guide which can be applied here or was applied to every page on the platform. Adding once again another form style will lead to more and more CSS lines and a quite bloated Stylesheet.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -14,6 +14,8 @@ class ApplicationController < ActionController::Base
 
   before_filter :set_geoip_cookie
 
+  before_filter :check_user_email!
+
   rescue_from Errno::ETIMEDOUT, :with => :timeout
   rescue_from Timeout::Error,   :with => :timeout
   rescue_from Net::ReadTimeout, :with => :timeout
@@ -36,6 +38,21 @@ class ApplicationController < ActionController::Base
 
   def default_url_options(options = nil)
     {:locale => I18n.locale}
+  end
+
+  def check_user_email!
+    # User logged in and has a blank email attribute?
+    if !current_user.blank? && current_user.email.blank?
+      current_user.errors.add_on_blank(:email)
+
+      respond_to do |format|
+        format.html{
+          redirect_to after_signup_edit_profile_path
+        }
+        format.json{render_exception(Exception.new(current_user.errors.full_messages.to_sentence), 403)}
+        format.xml{render_exception(Exception.new(current_user.errors.full_messages.to_sentence), 403)}
+      end
+    end
   end
 
   def authenticate_terms!
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -3,6 +3,7 @@ class UsersController < ApplicationController
   include AuthenticateUserFromToken
 
   skip_before_filter :verify_authenticity_token, :only => :authenticate
+  skip_before_filter :check_user_email!, only: [:after_signup_edit, :after_signup_update]
 
   before_filter :authenticate_user!,        :except => [:authenticate, :signed_in]
   before_filter :authenticate_mobile_user,  :only => :authenticate
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -72,7 +72,6 @@ class UsersController < ApplicationController
       end
       redirect_to after_sign_in_path_for(:user)
     else
-      flash.now[:alert] = @user.errors.full_messages.to_sentence
       render :template => 'users/after_signup_edit'
     end
   end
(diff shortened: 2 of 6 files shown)


lennerd:
I realized, that it was a total stupid idea to remove password completely as we have admin users who need a password to login. I will revert that part of the PR.

holgerd:
@lennerd I tried to test this on staging but get errors:

- Using Firefox Mac I click "login" and "login with OSM" and login to OSM using holger@mailinator.com. I get back to the add email form on wheelmap and type in holger@mailinator.com. I get a notification that I'll get a confirmation email and click "OK"
- I expect the then come to wheelmap homepage and be logged in
- Instead, I see a "502 bad gateway" error from cloudflare (see screenshot). When I reload the page https://staging.wheelmap.org/profile/after_signup_update (which is in the browser bar) I get "The page you were looking for doesn't exist."

staging_wheelmap_org___502__bad_gateway

lennerd:
This not on deployed on Staging so maybe you discovered another error.

holgerd:
ok, let me know when I can test something on staging.

Lennart Hildebrandt <notifications@github.com> schrieb am Di., 16. Aug.
2016 um 11:42 Uhr:

holgerd:
I described the bud in separate issue #361

lennerd:
👍 Thanks!

lennerd:
FYI: Holger and I discussed the flow again and agreed on just fixing the current "Almost there" view without making the users email address mandatory.

lennerd:
* [ ] Update new translations.

lennerd:
@1000miles What do you think about deploying this on Staging? Does it make sense to deploy this on Staging now? Or should we wait for the new server setup.

Xylakant:
go ahead and deploy on staging.

Xylakant:
there's no reason not to continue development as normal, we're building fully parallel

lennerd:
Didn't want to break something. Thanks for the information.

lennerd:
@holgerd @Svenyo This is deployed on staging. Looking forward to your feedback.

@Svenyo I removed a source translation string and would like to push this to Transifex to update all other translation files. Okay?

lennerd:
@holgerd @Svenyo Please review.

holgerd:
This is still blocked by #361

1000miles:
This may need to be deployed on staging again for @holgerd to review.

holgerd:
@1000miles can you look into this and decide whether this can be resolved?

holgerd:
@1000miles what is the status of this?

1000miles:
@holgerd The sign-up/sign-in part was put on hold until the front-end part got a make-over by @opyh ? This is why this PR (#359) and the related issues (#175, #179) were assigned to you, or?

holgerd:
@1000miles thanks. so this is on hold as well. :)

PR #376, opened 2016-09-05
Add new node types for leisure category.

lennerd:


1000miles:
Looks good to me, thank you. 🎉

holgerd:
It seems that the words "Massage spa" and "Tattoo shop" were used in all language versions and not just in English.

Also, I can not find "massage" in categories.yml in Transifex.

PR #382, opened 2016-09-06
Copy manifest to apps

Hoverbear:
@Xylakant I would appreciate a review on this, as your knowledge exceeds mine on this issue. :)

### DIFF
--- a/config/deploy.rb
+++ b/config/deploy.rb
@@ -80,12 +80,6 @@ namespace :deploy do
         remote_dir = "#{host.user}@#{host.hostname}:#{release_path}/public/assets/"
         execute "mkdir -p #{release_path}/public/assets/"
         run_locally { execute "rsync -av --delete #{local_dir}/.sprockets-manifest-* #{remote_dir}" }
-
-        # stolen from https://github.com/capistrano/capistrano/blob/master/lib/capistrano/tasks/deploy.rake#L101
-        # we can't directly invoke the task since that runs on deploy roles only
-        tmp_current_path = release_path.parent.join(current_path.basename)
-        execute :ln, "-s", release_path, tmp_current_path
-        execute :mv, tmp_current_path, current_path.parent
       end
 
       # clean up
--- a/config/deploy.rb
+++ b/config/deploy.rb
@@ -75,6 +75,19 @@ namespace :deploy do
         execute :mv, tmp_current_path, current_path.parent
       end
 
+      on roles(:app) do
+        # this needs to be done outside run_locally in order for host to exist
+        remote_dir = "#{host.user}@#{host.hostname}:#{release_path}/public/assets/"
+        execute "mkdir -p #{release_path}/public/assets/"
+        run_locally { execute "rsync -av --delete #{local_dir}/.sprockets-manifest-* #{remote_dir}" }
+
+        # stolen from https://github.com/capistrano/capistrano/blob/master/lib/capistrano/tasks/deploy.rake#L101
+        # we can't directly invoke the task since that runs on deploy roles only
+        tmp_current_path = release_path.parent.join(current_path.basename)
+        execute :ln, "-s", release_path, tmp_current_path
+        execute :mv, tmp_current_path, current_path.parent
+      end
+
       # clean up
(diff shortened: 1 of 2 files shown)


config/deploy.rb
```diff
@@ -75,6 +75,19 @@
...
       end
 
+      on roles(:app) do
+        # this needs to be done outside run_locally in order for host to exist
+        remote_dir = "#{host.user}@#{host.hostname}:#{release_path}/public/assets/"
+        execute "mkdir -p #{release_path}/public/assets/"
+        run_locally { execute "rsync -av --delete #{local_dir}/.sprockets-manifest-* #{remote_dir}" }
+
+        # stolen from https://github.com/capistrano/capistrano/blob/master/lib/capistrano/tasks/deploy.rake#L101
+        # we can't directly invoke the task since that runs on deploy roles only
```
Xylakant: app is a deploy role, so this code block should not be needed

config/deploy.rb
```diff
@@ -75,6 +75,19 @@
         execute :mv, tmp_current_path, current_path.parent
       end
 
+      on roles(:app) do
+        # this needs to be done outside run_locally in order for host to exist
+        remote_dir = "#{host.user}@#{host.hostname}:#{release_path}/public/assets/"
+        execute "mkdir -p #{release_path}/public/assets/"
+        run_locally { execute "rsync -av --delete #{local_dir}/.sprockets-manifest-* #{remote_dir}" }
```
Xylakant: you might also need the manifest.yml

Hoverbear: Hm, where would this `manifest.yml` be? 

```
wheelmap@asset:/var/apps/wheelmap/current$ find /var/apps/wheelmap/current/ | grep manifest
/var/apps/wheelmap/current/public/assets/.sprockets-manifest-ca4c4a76cef02a74bbae48b8ba49775d.json
```

Xylakant: i was having a  look at the current prod server to compare:

````
felix@spiderman:/var/apps/wheelmap/production/current/public/assets$ ls -la manifest.yml
-rw-rw-r-- 1 rails rails 80708 Jun  7 17:27 manifest.yml
````

config/deploy.rb
Hoverbear: Removed in ac13761

config/deploy.rb
Xylakant: yes, looks like the manifest.yml is a rails3 artefact.

Hoverbear: I manually rsynced a copy over and it seems to be working partially. 

screen shot 2016-09-06 at 14 26 09

Xylakant:
cool, can you do a test deployment?

Xylakant:
did you change your /etc/hosts to point to wheelmap to the staging lb?

Hoverbear:
Yes I changed my hostfile.

Xylakant:
the app on staging expects the hostname to be staging.wheelmap.org, not wheelmap.org

Hoverbear:
Ah that might be the reason.

Hoverbear:
I'm rebuilding the cluster now. Should I point the capistrano files at this branch or wait for merge before deploying the apps?

Xylakant:
why rebuild the cluster?

Hoverbear:
> cool, can you do a test deployment?

I was under the impression this was a full deployment end-to-end since there were changes in regards to consul as well as this app.

Xylakant:
we need to be able to roll out partial changes without rebuilding the whole infrastructure. We can't take down prod just because we changed a host name

Hoverbear:
https://github.com/sozialhelden/chef-repository/issues/121 will accomplish this.

Xylakant:
so currently all you need to do is run knife solo manually for all machines, but rebuilding shouldn't be needed.

Hoverbear:
**Please do not merge this until I remove the next commit, it is for testing deploy only**

Hoverbear:
This fixes all but some scss imports. I'll be making a seperate PR for this.

screen shot 2016-09-07 at 10 43 26

PR #391, opened 2016-09-08
Reflect the new node names. Worker has a number.

Hoverbear:
The worker nodes now have a number so we need to reflect this.


Hoverbear:
# Please don't merge this while the capistrano branch is changed.

Xylakant:
LGTM. please merge.

Issue #334, opened 2016-07-19
wheelchair-tag not shown on wheelmap, but toilet correctly shown


holgerd:
- I'm OSM-user nanonan and  and mark the place  http://wheelmap.org/nodes/-239534993 using wheelmap as wheelchair=yes and toilet=yes
- I expect both to become "green" and the data to be saved in OSM
- Instead, I the wheelchair-tag stays unmarked/grey and only the toilet tag changes to green. But my changes have been written to OSM as I can see in this changeset: http://www.openstreetmap.org/changeset/40795134

holgerd:
Svenja schreibt:

ja, ich konnte den Bug gerade mit der Web-App reproduzieren:

Der POI hatte einen grünen Zugänglichkeitsstatus, als ich die Toilette grün markierte, wurde der Zigänglichkeitsstatus grau. Dann auch das Icon grau. Wenn ich jetzt auf das Icon klicke, wird beides grün angezeigt. Wenn ich die Bubble schließe, ist das Icon komplett verschwunden.

Siehe Screenshots.
![bildschirmfoto 2016-07-18 um 10 59 19](https://cloud.githubusercontent.com/assets/218506/16939981/c0068868-4d86-11e6-9071-f24a3f12ad32.png)

![bildschirmfoto 2016-07-18 um 10 59 09](https://cloud.githubusercontent.com/assets/218506/16939984/c44851ae-4d86-11e6-830e-99c3e1a5c8cb.png)


![bildschirmfoto 2016-07-18 um 10 59 00](https://cloud.githubusercontent.com/assets/218506/16939989/cb54a15a-4d86-11e6-8a56-79dad47e8df1.png)

![bildschirmfoto 2016-07-18 um 10 58 46](https://cloud.githubusercontent.com/assets/218506/16939993/d0524626-4d86-11e6-852f-6a147b42c247.png)

holgerd:
@Svenyo Bei mir sieht der Freizeitpark ok aus. Ich habe eben danach gesucht in Wheelmap und auch gefunden und er ist grün/grün. http://wheelmap.org/nodes/1834315985

holgerd:
I notice the node-id in the wheelmap URL is negative, so I understand this is a POI which has been converted from a shape. Maybe this is related to this issue?

Svenyo:
Hej, true, seems the tag is displayed correctly after some time again.

holgerd:
happens on iOS as well; related?
[...]

PR #399, opened 2016-09-14
Wheelchair tag not shown on wheelmap

schultyy:
PR for #334 

This fixes the problem that a user could not update both toilet and wheelchair status for a POI at the same time. With this change it is now possible to update both at the same time.
Also some small refactorings took place in the `UpdateTagsJob` class like extracting magic strings and moving big conditionals into methods.

### DIFF
--- a/app/jobs/update_tags_job.rb
+++ b/app/jobs/update_tags_job.rb
@@ -58,19 +58,23 @@ class UpdateTagsJob < Struct.new(:element_id, :type, :tags, :user, :client, :sou
       api.save(element, changeset)
 
       Counter.increment(source)
-      if update_wheelchair?
-        user.increment!(:tag_counter) if user.terms?
-        update_poi!
-      elsif update_toilet?
-        user.increment!(:toilet_counter) if user.terms?
-        update_poi!
-      else
-        user.increment!(:edit_counter) if user.terms?
-      end
+      increment_user_counter!
     rescue Rosemary::Conflict => conflict
       logger.info "IGNORE: #{type}:#{element_id} nothing has changed! (conflict)"
       # These changes have already been made, so dismiss this update!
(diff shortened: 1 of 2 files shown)


app/jobs/update_tags_job.rb
```diff
@@ -1,15 +1,18 @@
...
+
   def self.enqueue(element_id, type, tags, user, source, tags_to_delete={})
     raise "user not app authorized" unless user.app_authorized? # implies user.access_token.present?
 
     # Do not enqeue job if not in production or test environment
     return unless Rails.env.production? || Rails.env.test?
 
     # Remove wheelchair tag if value is "unknown"
-    tags.delete("wheelchair") if tags["wheelchair"] == 'unknown'
+    tags.delete("wheelchair") if tags[WHEELCHAIR_TAG_KEY] == 'unknown'
```
1000miles: as discussed please change

1000miles: as discussed please change the code to following:
```ruby
tags.delete(WHEELCHAIR_TAG_KEY) if tags[WHEELCHAIR_TAG_KEY] == 'unknown'
```

app/jobs/update_tags_job.rb
```diff
@@ -1,15 +1,18 @@
...
 
     # Remove wheelchair tag if value is "unknown"
-    tags.delete("toilets:wheelchair") if tags["toilets:wheelchair"] == 'unknown'
+    tags.delete("toilets:wheelchair") if tags[WHEELCHAIR_TOILET_TAG_KEY] == 'unknown'
```
1000miles: also please change to:

```ruby
tags.delete(WHEELCHAIR_TOILET_TAG_KEY) if tags[WHEELCHAIR_TOILET_TAG_KEY] == 'unknown'
```

1000miles:
Thanks for the update! LGTM! 🍊

PR #412, opened 2016-10-04
Add ping file for healthchecks

Hoverbear:
Adds a ping file for asset host healthchecks.


Hoverbear:
I'm merging this since https://github.com/sozialhelden/chef-repository/issues/214 depends on this and was approved and merged. Thus if this is not merged and a re-deploy is done then the asset host will fail.

Issue #404, opened 2016-09-21
500 error when changing category of node_type


holgerd:
- I'm logged in as admin on https://wheelmap.org/admin/node_types/41/edit . I change the category of museum from "education" to "tourism" and click save
- I expect to see a success message
- Instead, I see "something went wrong" and the url is https://wheelmap.org/admin/node_types/41

1000miles:
Thanks for reporting this, @holgerd! :)

After a short research it throws same error as mentioned in #365 
`ActiveModel::ForbiddenAttributesError (ActiveModel::ForbiddenAttributesError)`

We will make sure to check all pages on the admin dashboard to fix this issue. Thanks for your patience.

1000miles:
@holgerd 

cc @schultyy 

So we have checked all pages on the active_admin dashboard of the production server and here are the results:

- `/admin/providers` => ERROR: `ActiveModel::ForbiddenAttributesError (ActiveModel::ForbiddenAttributesError)`  => Fixed in PR https://github.com/sozialhelden/wheelmap/pull/394
- `/admin/provided_pois` => ERROR: `ActiveModel::ForbiddenAttributesError (ActiveModel::ForbiddenAttributesError)`  => Fixed in PR https://github.com/sozialhelden/wheelmap/pull/394

----
- `/admin/addresses` => create,  edit & delete entry works!
- `/admin/admin_users` => create,  edit & delete entry works!
- `/admin/users`  => create,  edit & delete entry works!
- `/admin/categories` => ERROR: `ActiveModel::ForbiddenAttributesError (ActiveModel::ForbiddenAttributesError)` => Fix on the way
- `/admin/node_types` => ERROR: `ActiveModel::ForbiddenAttributesError (ActiveModel::ForbiddenAttributesError)`  => Fix on the way
- `/admin/photos` => ERROR:  `WARNING: Can't mass-assign protected` => Reported in Ticket #415
- `/admin/regions` => ERROR:  `ActionView::Template::Error` => Reported in Ticket #414

1000miles:
@holgerd The fix is deployed on staging and ready for review.

holgerd:
@1000miles  I tested this. It does not give me a 500 error but two things came up:

1. 
tourism exists twice on staging in the dropdown. I tested with both.

2. 
[...]

PR #416, opened 2016-10-04
Fix/forbidden attributes error in active admin

1000miles:
This PR fixes following error:
 `ActiveModel::ForbiddenAttributesError (ActiveModel::ForbiddenAttributesError)` 

in:
- /admin/categories 
- /admin/node_types

Related issue: #404

### DIFF
--- a/app/admin/category.rb
+++ b/app/admin/category.rb
@@ -6,11 +6,13 @@ ActiveAdmin.register Category do
     %i(identifier)
   end
 
-  # For some reason, inherited resource update does not work.
-  def update
-    region = resource
-    region.update_attributes(params[:category])
-    super
+  controller do
+    # For some reason, inherited resource update does not work.
+    def update
+      region = resource
+      region.update_attributes(params[:category][:id])
+      super
+    end
   end
 
   index do
--- a/app/admin/category.rb
+++ b/app/admin/category.rb
@@ -1,16 +1,17 @@
 ActiveAdmin.register Category do
 
-  controller do
+  filter :identifier
 
-    # For some reason, inherited resource update does not work.
-    def update
-      region = resource
-      region.update_attributes(params[:category])
-      super
-    end
+  permit_params do
+    %i(identifier)
   end
 
-  filter :identifier
+  # For some reason, inherited resource update does not work.
+  def update
+    region = resource
+    region.update_attributes(params[:category])
+    super
+  end
 
   index do
     selectable_column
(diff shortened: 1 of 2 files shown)


app/admin/category.rb
```diff
@@ -1,16 +1,17 @@
 ActiveAdmin.register Category do
 
-  controller do
```
schultyy: Why is `controller do` gone here?

1000miles: Ah, you are right, will fix the method `update_attributes` (because it also had `ForbiddenAttributes` error) instead of removing the controller. Thank you. :)

app/admin/category.rb
```diff
@@ -6,11 +6,13 @@
     %i(identifier)
   end
 
-  # For some reason, inherited resource update does not work.
-  def update
-    region = resource
-    region.update_attributes(params[:category])
-    super
+  controller do
```
Hoverbear: It looks like this was the syntax used in the original commit 4 years ago: https://github.com/sozialhelden/wheelmap/commit/f8b3b0a7afb8e71dc9e8f997f2520848fd2b0d55. Wasn't able to dig up where it was removed though.

So this looks good.

app/admin/category.rb
```diff
@@ -6,11 +6,13 @@
...
+    # For some reason, inherited resource update does not work.
+    def update
+      region = resource
+      region.update_attributes(params[:category][:id])
```
Hoverbear: Hm, did this change recently? 

I note [here](https://github.com/sozialhelden/wheelmap/blob/4f39534ef846a34eea1c6834779d37d123c29090/app/admin/users.rb#L33), [here](https://github.com/sozialhelden/wheelmap/blob/4f39534ef846a34eea1c6834779d37d123c29090/app/admin/poi.rb#L29), and [here](https://github.com/sozialhelden/wheelmap/blob/4f39534ef846a34eea1c6834779d37d123c29090/app/admin/region.rb#L14) all use the parameter, not an ID.

1000miles: Thanks for the hint. According to active_admin's docs we need to use `permitted_params` inside controller actions.

```
The permit_params call creates a method called permitted_params. You should use this method when overriding create or update actions:

ActiveAdmin.register Post do
  controller do
    def create
      # Good
      @post = Post.new(permitted_params[:post])
      # Bad
      @post = Post.new(params[:post])

      if @post.save
        # ...
      end
    end
  end
end
```

Hoverbear: 👍

1000miles:
Thanks for the review. :)

PR #427, opened 2016-10-07
[WIP] Purge `public/` of Riff-Raff

### LINKED ISSUE #424: Migrate non-critical things from `public/`
Currently many random files are served from `public/`. Some have a clear purpose and are obviously of importance, others are less so.

As part of our infra-rebuild we'd like to minimize/clean out this folder.

You can review the current contents [here](https://github.com/sozialhelden/wheelmap/tree/master/public).

This issue will be updated with plans for each item.

Hoverbear:
Do some cleaning on the `public/` folder.

Closes #424.


Hoverbear:
So this should be ready for merge. It's deployed on staging and is producing no issues as far as I can tell.

Hoverbear:
Thanks @lennerd :) We actually think the icons.zips might be able to removed from here as well since there seems to be another place they are updated. 

Going to leave them in this commit since I don't want to break anything. :)

lennerd:
Did you check the asset API endpoint? If the linked URIs point to the `asset` folder inside `app` you should be totally safe to remove them from `public`. cc @holgerd

Hoverbear:
I'll check that before merging. :)

Hoverbear:
`public/icons/` are used on the admin area node_types screen, can't purge them.

lennerd:
Does it make sense to change that? I mean the correct asset path is available anyway.

Hoverbear:
Blocked regarding icons/marker.zip

Hoverbear:
I reverted the commit that removed the `.zip` files due to https://github.com/sozialhelden/wheelmap/issues/430#issuecomment-272885574 .

1000miles:
Talked with @hoverbear today. 

### Please don't merge, we are still not 100% sure about all files to be removed. This needs way more investigations which will take time.

Issue #413, opened 2016-10-04
Enhance nodes API to accept measurements generated by Project Tango


schultyy:
As requested by SMF, the nodes API should be enhanced in a way so measurements can be stored for each node.

Pasting the proposal from SMF here:

```
Messungsbild hochladen:

/api/nodes/{node_id}/measurement
POST (wie /api/nodes/{node_id}/photos)

Response:
{"message":"OK", "id": 123456}

===========================================

Metadaten hochladen:

/api/nodes/{node_id}/measurement/{id}/meta
PUT oder POST

{
        'type': 'DOOR/STEPS/RAMP/TOILET',
        'description': 'some user description',
        'data': {
                // daten abhängig von dem typ
                // DOOR
                'width': 1.04 //meter

                // STEPS
                'height': 0.11 //meter

                // RAMP
                'angle': 15.42 // degree

                // TOILET
                'space': 2.04 // m^2

        }
}
```

A measurement consists for an image (Messungsbild) and meta data such as:
- width (in meters)
- height (in meters)
- angle (in degrees)
- space (in m^2)

schultyy:
After phone calls with Raphael Heber, Tim Freiheit and Holger Dietrich the following was agreed upon:

The first version if this feature will include the following:

- Two new API endpoints which will allow creating new measurement records (details below)
- Measurement photos for a node will be returned together with the other regular photos for that node.

It will not be possible to fetch meta information for a measurement yet. One option here would be to provide these via another endpoint. All apps (iOS, Android, Web, Windows Phone) then need to talk to that endpoint as well. This solution is not feasible right now due to budget reasons.

The other option would be to enhance the existing endpoint which delivers photos for a node. This could possibly break existing installations on Android, iOS, Windows Phone and also the web application.

## Create new measurement

There will be two endpoints handling creation of measurements.

1. Upload measurement image

```
POST /api/nodes/:nodeid/measurement
(image is sent as multipart)

=> 201 Created
{"id": 4234523543}
```

This will return `201 Created` with the new image id if successful.

2. Upload metadata
Upload metadata for an existing measurement image.

```
POST /api/nodes/:nodeid/measurement/:id/metadata
{
        'type': 'DOOR/STEPS/RAMP/TOILET',
        'description': 'some user description',
        'data': {
                // Data depends on selected type
                // DOOR
                'width': 1.04 //meter
                // STEPS
                'height': 0.11 //meter
                // RAMP
                'angle': 15.42 // degree
[...]

PR #428, opened 2016-10-07
Store measurement photos

schultyy:
PR for #413 

This PR adds the actual implementation behind:
```
POST /api/nodes/:node_id/measurements
```

These measurement images are stored as regular images. With this change we also ensure that they're going to be displayed together with the other regular images as well.

### DIFF
--- a/spec/controllers/api/measurements_controller_spec.rb
+++ b/spec/controllers/api/measurements_controller_spec.rb
@@ -89,6 +89,20 @@ describe Api::MeasurementsController do
           expect(response.status).to eq 401
         end
       end
+
+      context 'with non existant node' do
+        before do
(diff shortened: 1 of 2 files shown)


spec/controllers/api/measurements_controller_spec.rb
```diff
@@ -56,6 +70,38 @@ def json_response
...
+        it 'does not save a new photo for the poi' do
+          expect(poi.photos.count).to eq 0
+        end
+
+        it 'returns 401 Unauthorized' do
+          expect(response.status).to eq 401
+        end
+      end
+
+      context 'with non existant node' do
```
1000miles: `existant` > `existent` 

just a small typo but since it does not change anything we can change it in the next round.

PR #431, opened 2016-10-11
Add a Vagrant box for easier development

Hoverbear:
We want to make sure new contributors have an easy time setting up to both run and develop on Wheelmap. 

This PR is based off https://github.com/Asquera/wheelmap-devenv and expands on its scope to be a comprehensive development environment setup solution.

It:
- [x] Forwards the SSH agent.
- [x] Installs all requirements (including bundle and node deps).
- [x] Configures the database.
- [x] Imports the Berlin dataset.
- [x] Mounts using `nfs` to avoid the problems with VirtualBox syncing.

You can test this by taking this branch and running `vagrant up` then `cd /vagrant && bundle exec rails server` in the box. Navigating to `localhost:3000` after should yield a working Wheelmap instance.


Vagrantfile
```diff
@@ -0,0 +1,99 @@
...
+  # Sync the folder in a way that isn't so slow with rails.
+  # Sync with `vagrant rsync`.
+  config.vm.synced_folder ".", "/vagrant", type: "rsync",
+    rsync__exclude: ["vendor/bundle/", "node_modules/"]
+
+  # Forwards the SSH agent so your keys work.
+  config.ssh.forward_agent = true
+
+  # Copy over the user's ssh config.
+  config.vm.provision "file", source: "~/.ssh/config", destination: ".ssh/config"
```
Xylakant: normal users won't do a deployment, so they don't need this.

Vagrantfile
```diff
@@ -0,0 +1,99 @@
...
+
+Vagrant.configure("2") do |config|
+  config.vm.box = "ubuntu/xenial64"
+
+  # Expose a port so the application can be tested.
+  config.vm.network "forwarded_port", guest: 3000, host: 3000
+
+  # Sync the folder in a way that isn't so slow with rails.
+  # Sync with `vagrant rsync`.
+  config.vm.synced_folder ".", "/vagrant", type: "rsync",
```
Xylakant: i don't think using rsync is a good idea. it's a one-time sync at startup, many users will expect changes to show up inside the machine without a restart.

relevant doc link: https://www.vagrantup.com/docs/synced-folders/rsync.html
````
The rsync synced folder does a one-time one-way sync from the machine running to the machine being started by Vagrant.
````

Comment (Hoverbear) [inline Vagrantfile]: Right, as mentioned in the README.md change:

> You can use `vagrant rsync` to sync changes between your local and box copies.

I agree that it's not ideal, but with the Virtualbox syncing slowness (it's really, really slow) and NFS being rather unreliable it seemed like the best option.

I can switch to to NFS and see if I can make it more reliably succeed.

File: Vagrantfile (hunk 1 above)
Comment (Hoverbear) [inline Vagrantfile]: Good point. :)

File: Vagrantfile (hunk 2 above)
Comment (Xylakant) [inline Vagrantfile]: so I do a one-line change, then call vagrant sync and then refresh in the browser? I'd go for virtualbox by default.

Comment (Hoverbear) [inline Vagrantfile]: This is the reason why:

```
ubuntu@ubuntu-xenial:/vagrant$ time bundle exec rake db:create:all
wheelmap_development already exists
wheelmap_development already exists

real	7m19.381s
user	0m4.232s
sys	3m11.144s
```

It takes 7+ minutes just to start `rake` using the Virtualbox mount.

Comment (Xylakant) [inline Vagrantfile]: i'd really investigate if that's not a problem of your setup. the virtualbox fs is known to be slow, but this looks way too slow.

Comment (Hoverbear) [inline Vagrantfile]: This is a known issue from the old repo https://github.com/1000miles/wheelmap-devenv/blob/8e9bda06583343b0c65e8d8e28ec929694a4315e/README.md#ubuntu--nfs-mount which @1000miles  solved with `nfs`.

Comment (Xylakant) [inline Vagrantfile]: i know that it's slow, but 7 minutes is way too slow (btw: you're measuring connectivity to the database as well here)

Comment (Hoverbear) [inline Vagrantfile]: This is the error I get with `nfs`:

```
==> default: /var/lib/gems/2.3.0/gems/bundler-1.13.3/lib/bundler/definition.rb:179:in `rescue in specs': Your bundle is locked to activesupport (4.1.15), but that version could not be found in any of the sources listed in your Gemfile. If you haven't changed sources, that means the author of activesupport (4.1.15) has removed it. You'll need to update your bundle to a different version of activesupport (4.1.15) that hasn't been removed in order to install. (Bundler::GemNotFound)
```

As a side note. Doesn't occur with other schemes.

**Comment by Hoverbear** [comment]
I worked out the nfs kinks so this should be good to go. :)

**Comment by Hoverbear** [comment]
I tested with Vagrant sharing again today on a new machine and entirely new setup. Still really slow to start:

```
tubuntu@ubuntu-xenial:/vagrant$ time bundle exec rails server
=> Booting Thin
=> Rails 4.1.15 application starting in development on http://0.0.0.0:3000
=> Run `rails server -h` for more startup options
=> Notice: server is listening on all interfaces (0.0.0.0). Consider using 127.0.0.1 (--binding option)
=> Ctrl-C to shutdown server
^CExiting
^C
real	3m46.853s
user	0m3.712s
sys	1m47.744s
```

Hoverbear:
So I moved the `bundle install` to a folder outside of the share and it seems to be working quite well. This is ready for review/testing @1000miles :)

Hoverbear:
@emig I'm reverting some of your changes because they were not part of the PR which was approved.

Issue #433, opened 2016-10-12
Export POI data as CSV-file from the backend for Streetspotr


Svenyo:
For the new project with our crowdsourcing partner Streetspotr http://news.wheelmap.org/streetspotr-brechen-alle-rekorde-28-000-bilder-gesammelt/ we need an export of specific POI data into a csv-file.

**The cities where those POIs are located are:**
Italy: Mailand, Rom, Neapel
Spain: Madrid, Barcelona, Valencia
France: Paris, Marseille, Lyon
UK: London, Birmingham, Glasgow
Finland: Helsinki, Tampere, Turku
Denmark: Kopenhagen, Aarhus, Odense
Sweden: Stockholm, Göteborg (Goeteborg?), Malmö (Malmoe?)
Netherlands: Amsterdam, Rotterdam, Den Haag
Belgien: Antwerpen, Charleroi, Brüssel (Bruessel?)
Irland: Dublin, Cork, Galway
Russia: Moskau, Sankt Petersburg, Nowosibirsk
Rumänien: Bukarest, Klausenburg, Temeswar

Deutschland (the 20 biggest cities): Berlin, Hamburg, Muenchen, Koeln, Frankfurt am Main, Stuttgart, Duesseldorf, Dortmund, Essen, Leipzig, Bremen, Dresden, Hannover, Nuernberg, Duisburg, Bochum, Wuppertal, Bielefeld, Bonn, Münster (sic! Here the "Umlaut" is available.)
1. Svenja has checked which cities are not available as regions in the backend. The following need to be imported into our backend first:
   https://docs.google.com/spreadsheets/d/13VpThehPWJ-y_ZdoXMKw3UVnsJ8PrrIPblR6bd0jpkQ/edit?usp=sharing 
2. @holgerd is trying to get the currently missing, necessary WKT-files for the shapes of the cities and upload them here when available.
3. These WKT-files need to be implemented into our backend.
4. After new regions have been defined in the backend, we need an export of POI-data in a csv format. Each country should have an own sheet with the data of the 3 (or for Germany: 20) cities. Example: The CSV file for Italy should contain all POIs of the needed categories, accessibility status and photo parameter for Mailand, Rom and Neapel.

As I just checked, I cannot do an export using the UI of the current backend because the filter settings are not working for creating a corresponding csv sheet (see: https://github.com/sozialhelden/wheelmap/issues/432)
@christoph-buente had a certain Script for that which should be documented.
[...]

PR #438, opened 2016-10-14
Import 31 new regions

1000miles:
This PR imports new regions and fixes parts of #433.

### Please merge #443 first, so we can rebase and use the umlauts fix in this import


lib/tasks/region.rake
```diff
@@ -44,7 +44,7 @@ namespace :region do
 
   desc 'Create a region for the hole world'
```
Hoverbear: (Unrelated to your PR) This should be 'whole' :)

PR #445, opened 2016-10-19
Fix uploads on new infra

Hoverbear:
Due to permissions and the NFS mounts on the new infra we need a bit of a different way of handling uploads.

Tested on new production.

screen shot 2016-10-19 at 16 08 44

### DIFF
--- a/config/deploy.rb
+++ b/config/deploy.rb
@@ -73,6 +73,11 @@ namespace :deploy do
         remote_dir = "#{host.user}@#{host.hostname}:#{release_path}/public/"
         execute "mkdir -p #{release_path}/public/assets/"
         run_locally { execute "rsync -av --delete #{local_dir} #{remote_dir}" }
+
+        # We need to link the existing shared `/var/apps/wheelmap/public/system` folder
+        # into the new release.
+        execute :ln, "-s", "#{deploy_path}/public/system", "#{release_path}/public/system"
+
         # We create this file so the consul health check will pass. We can't use an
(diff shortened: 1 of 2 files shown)


config/deploy.rb
```diff
@@ -73,6 +73,11 @@
         remote_dir = "#{host.user}@#{host.hostname}:#{release_path}/public/"
         execute "mkdir -p #{release_path}/public/assets/"
         run_locally { execute "rsync -av --delete #{local_dir} #{remote_dir}" }
+
+        # We need to link the existing shared `/var/apps/wheelmap/public/system` folder
+        # into the new release.
+        execute :ln, "-s", "#{deploy_path}/public/system/uploads", "#{release_path}/public/system"
```
schultyy: Isn't this done automatically by Capistrano?

Hoverbear: If I recall this is because we no-op `set_linked_dirs`

```ruby
task :set_linked_dirs do
      # noop due to cap/rails adding this to the run list
      # Ref: https://github.com/capistrano/rails/blob/54da36a2dda5084a3cbe380b4a8f1ba282379f72/lib/capistrano/tasks/assets.rake#L121
end
```

Xylakant: @Hoverbear is right. Since we do the build on the deploying machine and not the target machine and we only upload everything we want and need, we can't use the default linked_dirs task which tries to link directories that don't exist. So we noop that task and manually link just the ones we want.

Issue #82, opened 2015-12-08
Umlauts lead to duplicate wkt-entries


thermann78:
We have a problem with importing new regions. As by now, regions are imported as .wkt-files into https://github.com/sozialhelden/wheelmap/tree/master/db/data/wkt/.
Have a close look to regions with umlauts in their name: in folder /europe/austria, /europe/germany and /europe/switzerland, their are several duplicates like "baden_wuerttemberg.wkt" and "baden_wurttemberg.wkt" as well as duplicate folder names.
This leads to problems with our Librato metrics, where we have to reassign these data sources to the corresponding dashboards.
As we plan to expand and add new regions soon this should be fixed very soon.
@lennerd did the import of the .wkt-files from a Dropbox-folder and startet the import task so he can tell how exactly new regions are added to the server.

thermann78:
@holgerd : This should be fixed within one of the next milestones  as it leads to problems every time when we import new regions that are important for administrative purposes - as happened now again: https://github.com/sozialhelden/wheelmap-privateissues/issues/29

1000miles:
Thank you for reporting this. 

**Here is our additional input:**
Mac OS x seems to have problems to handle 2 directories with the same name but with different encoding.

**Example:**
On Mac OS we see 2x `Baden-Württemberg` folders in the directory: 
*wheelmap/db/data/wkt/Europe/Germany/*

```
→ ls
Baden_Württemberg
Baden_Württemberg
```
**When we try to access both of them, we can only access 1 of them:**
```
→ cd Baden_Württemberg/
Bodenseekreis/   Kreis_Konstanz/
```
On the virtual machine (OS: Ubuntu) it is possible to see and access both `Baden-Württemberg` folders by using the `tab`key to switch between them.

**First `Baden_Württemberg` folder is selected:**
```
➜  Germany git:(add-new-regions-germany-mexico) cd Baden_Wu<0308>rttemberg/
Baden_Württemberg/  Baden_Württemberg/
```
**Second `Baden_Württemberg` folder is selected:**
```
➜  Germany git:(add-new-regions-germany-mexico) cd Baden_Württemberg/
Baden_Württemberg/  Baden_Württemberg/
```

There is also an issue when using Git because it recognizes that there are 2 folders with the same name but with different encoding. When we try to change one of them, both are marked as changed for Git. When we try to delete 1 of them, we are not able to catch it, because of the encoding issue.

Xylakant:
[...]

PR #447, opened 2016-10-19
Find resource by slug instead of name

1000miles:
This PR fixes the `500 Internal Server Error` in `/admin/regions` after some wkt-files have been renamed due to umlauts issue (https://github.com/sozialhelden/wheelmap/issues/82). 

Finding a resource by `slug` is recommended rather than finding by `name` because names are supposed to change more often.

Example:
```
+-----+------------------+------------------+
| id  | name             | slug             |
+-----+------------------+------------------+
| 000 | Kreis Böblingen | kreis-boeblingen | 
+-----+------------------+------------------+
1 row in set (0.00 sec)
```

### DIFF
--- a/app/admin/region.rb
+++ b/app/admin/region.rb
@@ -6,7 +6,7 @@ ActiveAdmin.register Region do
 
   controller do
     def find_resource
-      Region.find_by(name: params[:id])
+      Region.find_by(slug: params[:id])
     end
 
     def update


1000miles:
Related PR: https://github.com/sozialhelden/wheelmap/pull/443

schultyy:
I do have a small nitpick here:
Using a slug over a name is preferred here because slugs are URL-safe where names are not. A name can contain letters which are not part of the ASCII charset like Umlauts or Cyrillic characters. 
A slug is a stripped down version of the name cleaned from all non-ASCII letters and also does not contain any white space.

PR #452, opened 2016-10-20
Only log WARN and ERROR on production.

Hoverbear:
This will help reduce log volumes.


schultyy:
Just gave it a thought if we may lose information here but the actual request is also logged on the load balancer.

config/environments/production.rb
```diff
@@ -40,7 +40,7 @@
   # config.force_ssl = true
 
   # See everything in the log (default is :info)
```
schultyy: Small nitpick: I guess this comment is obsolete.

PR #458, opened 2016-10-24
Fix querying for Pois without images

schultyy:
This improves the total time the query runs. For Berlin, Hamburg, Munich it took over 20 mins before, now it takes around 5.
We updated the mechanism for querying for Pois which do not have any photos associated. Also we sort node type ids before we pass them in.


1000miles:
Thanks for the PR! 🚀 Merged. 

My best,
Thuy

Issue #444, opened 2016-10-19
Running `bundle exec rake poi:locate` fails on staging and locally


1000miles:
At the moment running `RAILS_ENV=staging bundle exec rake poi:locate` manually and directly on the staging and locally fails.

**Error output on staging server**
```
xxx@xxxx:/var/apps/wheelmap/staging/current$ RAILS_ENV=staging bundle exec rake poi:locate
rake aborted!
NoMethodError: undefined method `grenze' for [:conditions, {:parent_id=>nil}]:Array
/var/apps/wheelmap/staging/releases/hidden_number/lib/poi_locator.rb:114:in `block in load_regions'
/var/apps/wheelmap/staging/releases/hidden_number/lib/poi_locator.rb:111:in `each'
/var/apps/wheelmap/staging/releases/hidden_number/lib/poi_locator.rb:111:in `load_regions'
/var/apps/wheelmap/staging/releases/hidden_number/lib/poi_locator.rb:14:in `initialize'
/var/apps/wheelmap/staging/releases/hidden_number/lib/tasks/poilocator.rake:7:in `new'
/var/apps/wheelmap/staging/releases/hidden_number/lib/tasks/poilocator.rake:7:in `block (2 levels) in <top (required)>'
Tasks: TOP => poi:locate
```

Christoph wrote:
>Looks like the DB schema is not up2date

----

**Error output locally (on a separate & master branch)**
```
vagrant@wheelmap:~/wheelmap$ RAILS_ENV=development bundle exec rake poi:locate
rake aborted!
NameError: uninitialized constant RGeo::Geos::Factory
/home/vagrant/wheelmap/lib/poi_locator.rb:12:in `initialize'
/home/vagrant/wheelmap/lib/tasks/poilocator.rake:7:in `new'
/home/vagrant/wheelmap/lib/tasks/poilocator.rake:7:in `block (2 levels) in <top (required)>'
Tasks: TOP => poi:locate
(See full trace by running task with --trace)
```

1000miles:
@Hoverbear Is this issue maybe related to your PR https://github.com/sozialhelden/wheelmap/pull/358?

This is the code part that raises the error:
```
  # load regions from database
  def load_regions(parent_id = nil)
    rr = Array.new
    Region.parent_id(parent_id).each do |db_region|
      r = Hash.new

      r[:geometry]   = db_region.grenze
      r[:id]         = db_region.id
      r[:name]       = db_region.name
      rr.push(r)
    end
[...]

PR #461, opened 2016-10-26
Update required packages section

schultyy:
We need a couple more packages in order to be able to run poi locate tasks properly


1000miles:
Related issue:
https://github.com/sozialhelden/wheelmap/issues/444

PR #470, opened 2016-11-02
Switch from git to https protocol for carrierwave source

1000miles:
This PR switches the source from `git` to `https protocol` for carrierwave.

Related warning while `bundle install` process:
```
The git source `git://github.com/carrierwaveuploader/carrierwave.git` uses the `git` protocol, which transmits data without encryption. Disable this warning with `bundle config git.allow_insecure true`, or switch to the `https` protocol to keep your data secure.
```


1000miles:
Thanks for the review! :)

PR #474, opened 2016-11-03
Pin ActiveAdmin to version 1.0.0.pre2 for now

schultyy:
This PR pins activeadmin at 1.0.0.pre2 to prevent breaking changes for later versions.

We figured out that if we use a more recent version it breaks a lot of stuff right now. Until we've time to investigate we stick with pre2.


1000miles:
sänks for the poool request! :)

PR #481, opened 2016-11-24
Remove German/English locale restriction in header menu

### LINKED ISSUE #479: [Translations] Locale Switcher does not apply to header > navigation
**Preconditions:**
- we look at the `navigation` in the header 
- we check the transifex key-value pair of `projects: Get Involved` that has been changed from `projects: Projects` in the resource file [relaunch.yml](https://github.com/sozialhelden/wheelmap/pull/475/files#diff-d250e923a9a2d44e6e304a690ffaa755) (EN)
- we compare the english relaunch.yml file with the spanish [relaunch.yml](https://github.com/sozialhelden/wheelmap/pull/475/files#diff-b22223ea41ce8d94668e909ce78923e6) and see that `projects: Proyectos` has also changed to `projects: Involúcrate` which is correct
- we export the newest translations with `bundle exec rake i18n:js:export` and can see them modified in `app/assets/javascripts/i18n` which is correct
- we check the spanish javascript file [es.js](https://github.com/sozialhelden/wheelmap/pull/475/files#diff-1bbd4dfc4e71abab8ef1e184e0297bfe) and can confirm that also here the key-value-pair has changed from `projects: Proyectos` to `projects: Involúcrate` which is correct

**Now:**
- we go to `https://staging.wheelmap.org` and switch to `german` and get the value `Mitmachen` for `projects` => correct!
- we go to `https://staging.wheelmap.org` and switch to `english` and get the value `Get Involved` for `projects` => correct!
- we go to `https://staging.wheelmap.org` and switch to `french` but get the value `Get Involved` for `projects` => not correct!
- we go to `https://staging.wheelmap.org` and switch to `spanish` but get the value `Get Involved` for `projects` => not correct!
- this goes on with the other languages...

**Conclusion:** 
- it seems the locale switcher does not work for the navigation neither locally, nor `staging` and `production` server
- for example on wheelmap.org when switching to spanish it should display `Proyectos` (old value) but it still shows `Projects` (old value) => https://github.com/sozialhelden/wheelmap/blob/master/app/assets/javascripts/i18n/es.js#L769
- no matter what language we switch to the same behavior applies to the others as well

schultyy:
This change enables the user to see the header items localized in their
language as well. Before it contained a lock on either German or
English.

Resolves #479

### DIFF
--- a/app/views/relaunch/_header.html.haml
+++ b/app/views/relaunch/_header.html.haml
@@ -40,9 +40,9 @@
         .nav-collapse.collapse
           .list-wrapper
             %ul.inline
-              %li= link_to t('header.navigation.about_wheelmap', :default => :en, :locale => (I18n.locale == :de ? :de : :en)), community_about_url
-              %li= link_to t('header.navigation.map', :default => :en, :locale => (I18n.locale == :de ? :de : :en)), root_path
-              %li= link_to t('header.navigation.projects', :default => :en, :locale => (I18n.locale == :de ? :de : :en)), community_projects_url
-              %li= link_to t('header.navigation.blog', :default => :en, :locale => (I18n.locale == :de ? :de : :en)), community_blog_url
-              %li= link_to t('header.navigation.press', :default => :en, :locale => (I18n.locale == :de ? :de : :en)), community_press_url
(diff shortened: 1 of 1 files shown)


1000miles:
LGTM, seal of approval. 👍  Thanks! :)

Issue #484, opened 2016-11-25
[Icon] Calling icon `art_museum.png` on node_types page results in 404 Error


1000miles:
This issue is a follow-up of ticket #476 

At the moment we get a `404 error` when calling the icon `http://staging.wheelmap.org/images/icons/art_museum.png` that is listed on the [node_types](http://staging.wheelmap.org/node_types) page.

Looking into the database we see that the icon `art_museum.png` still exists while this icon is not referenced anywhere and not existent in the app (anymore). It might have been removed but was not cleaned up in the database. 

```
mysql> SELECT * FROM node_types WHERE icon LIKE "%museum%";
+-----+-------------+---------------------+---------+---------------------+--------------------------+---------------------+---------------------+-------------+---------------+
| id  | category_id | identifier          | osm_key | osm_value           | icon                     | created_at          | updated_at          | alt_osm_key | alt_osm_value |
+-----+-------------+---------------------+---------+---------------------+--------------------------+---------------------+---------------------+-------------+---------------+
|  26 |           3 | gallery             | tourism | gallery             | museum_art.png           | 2012-10-31 10:55:01 | 2013-02-06 14:41:35 | amenity     | gallery       |
| 153 |           6 | art                 | shop    | art                 | art_museum.png           | 2015-12-01 22:36:07 | 2015-12-01 22:36:07 | NULL        | NULL          |
| 157 |           6 | antiques            | shop    | antiques            | art_museum_2.png         | 2015-12-01 22:36:07 | 2015-12-01 22:36:07 | NULL        | NULL          |
+-----+-------------+---------------------+---------+---------------------+--------------------------+---------------------+---------------------+-------------+---------------+
6 rows in set (0.00 sec)
```

Task:
- rename icon `art_museum.png` > `museum_art.png`

1000miles:
@karolinamakaruk cc @holgerd 

This issue is fixed and deployed on staging with release [staging-20161128143648](https://github.com/sozialhelden/wheelmap/releases/tag/staging-20161128143648). Ready for review. :)

<img width="1276" alt="00000202" src="https://cloud.githubusercontent.com/assets/4077916/20675358/4aa0c35e-b58c-11e6-8467-47f2b22cd5f1.png">

1000miles:
Closed. Deployed on production with release: https://github.com/sozialhelden/wheelmap/releases/tag/production-20161129150628.

PR #486, opened 2016-11-28
Rename and migrate art_museum.png to museum_art.png in node_types table

1000miles:
This PR fixes `404 - icon not found` error for "Art Museum"

Related Github issue: #484

### DIFF
--- a/db/migrate/20161128102235_change_art_museum_to_museum_art.rb
+++ b/db/migrate/20161128102235_change_art_museum_to_museum_art.rb
@@ -10,8 +10,6 @@ class ChangeArtMuseumToMuseumArt < ActiveRecord::Migration
     unless node_type.nil? && node_type.icon == 'art_museum.png'
       node_type.icon = 'museum_art.png'
       node_type.save
-    else
-      return
     end
   end
 
@@ -20,8 +18,6 @@ class ChangeArtMuseumToMuseumArt < ActiveRecord::Migration
     unless node_type.nil? && node_type.icon == 'museum_art.png'
       node_type.icon = 'art_museum.png'
       node_type.save
-    else
-      return
     end
   end
 end
--- a/db/migrate/20161128102235_change_art_museum_to_museum_art.rb
+++ b/db/migrate/20161128102235_change_art_museum_to_museum_art.rb
@@ -1,0 +1,27 @@
+class ChangeArtMuseumToMuseumArt < ActiveRecord::Migration
+  class LocalNodeType < ActiveRecord::Base
+    self.table_name = 'node_types'
+    validates :icon, presence: true
+  end
+
+  # Rename art_museum.png to museum_art.png but keep same identifier & osm_value
+  def up
+    node_type = LocalNodeType.find_by(id: 153)
+    if node_type.nil?
+      return
+    else
+      node_type.icon = 'museum_art.png'
+      node_type.save
+    end
+  end
+
+  def down
+    node_type = LocalNodeType.find_by(id: 153)
+    if node_type.nil?
+      return
+    else
+      node_type.icon = 'art_museum.png'
+      node_type.save
+    end
+  end
+end
--- a/db/migrate/20161128102235_change_art_museum_to_museum_art.rb
+++ b/db/migrate/20161128102235_change_art_museum_to_museum_art.rb
@@ -1,27 +1,27 @@
 class ChangeArtMuseumToMuseumArt < ActiveRecord::Migration
   class LocalNodeType < ActiveRecord::Base
     self.table_name = 'node_types'
-    validates :icon, presence: true
+    validates :identifier, :osm_value, :icon, presence: true
   end
 
   # Rename art_museum.png to museum_art.png but keep same identifier & osm_value
   def up
-    node_type = LocalNodeType.find_by(id: 153)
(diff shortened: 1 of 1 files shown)


db/migrate/20161128102235_change_art_museum_to_museum_art.rb
```diff
@@ -0,0 +1,27 @@
+class ChangeArtMuseumToMuseumArt < ActiveRecord::Migration
+  class LocalNodeType < ActiveRecord::Base
+    self.table_name = 'node_types'
+    validates :icon, presence: true
+  end
+
+  # Rename art_museum.png to museum_art.png but keep same identifier & osm_value
+  def up
+    node_type = LocalNodeType.find_by(id: 153)
```
schultyy: I'm currently thinking if it's the best way to load by `id` because there's the possibility that on Staging the id is 153, but on production for example it's 155. My advice is to load the node type by its `identifier` and `osm_key`. You still can be sure you get the record you want but you do not need to rely on the id to be the same across all environments.

db/migrate/20161128102235_change_art_museum_to_museum_art.rb
```diff
@@ -0,0 +1,27 @@
...
+    if node_type.nil?
+      return
+    else
+      node_type.icon = 'museum_art.png'
```
schultyy: Right now you just overwrite the `node_type`'s `icon`with the new value. What if that `node_type` already has the right icon? To be safe here what about checking beforehand if `node_type.icon` has the expected wrong value (`art_museum.png`)?

db/migrate/20161128102235_change_art_museum_to_museum_art.rb
```diff
@@ -0,0 +1,27 @@
...
+      node_type.save
+    end
+  end
+
+  def down
+    node_type = LocalNodeType.find_by(id: 153)
```
schultyy: Same as above.

db/migrate/20161128102235_change_art_museum_to_museum_art.rb
```diff
@@ -0,0 +1,27 @@
...
+    if node_type.nil?
+      return
+    else
+      node_type.icon = 'art_museum.png'
```
schultyy: Same as above but check for `museum_art.png`.

db/migrate/20161128102235_change_art_museum_to_museum_art.rb
1000miles: Thanks for pointing this out. You are right, maybe it is much better to chain the query like

search for `museum_art.png` (existent)
```
2.2.2 :010 > node_type = NodeType.find_by(identifier: "art", osm_value: "art", icon: "museum_art.png")
14:36:58   NodeType Load (0.7ms)  SELECT  `node_types`.* FROM `node_types`  WHERE `node_types`.`identifier` = 'art' AND `node_types`.`osm_value` = 'art' AND `node_types`.`icon` = 'museum_art.png' LIMIT 1
14:36:58  => #<NodeType id: 153, category_id: 3, identifier: "art", osm_key: "shop", osm_value: "art", icon: "museum_art.png", created_at: "2016-11-28 10:15:05", updated_at: "2016-11-28 11:30:06", alt_osm_key: "NULL", alt_osm_value: "NULL">
```

search for `art_museum.png` (non-existent):
```
2.2.2 :012 > node_type = NodeType.find_by(identifier: "art", osm_value: "art", icon: "art_museum.png")
  NodeType Load (1.2ms)  SELECT  `node_types`.* FROM `node_types`  WHERE `node_types`.`identifier` = 'art' AND `node_types`.`osm_value` = 'art' AND `node_types`.`icon` = 'art_museum.png' LIMIT 1
 => nil
```

I will specify this and add the changes to the PR.

db/migrate/20161128102235_change_art_museum_to_museum_art.rb
1000miles: I see. Thanks, will update the PR.

db/migrate/20161128102235_change_art_museum_to_museum_art.rb
```diff
@@ -0,0 +1,27 @@
...
+  end
+
+  # Rename art_museum.png to museum_art.png but keep same identifier & osm_value
+  def up
+    node_type = LocalNodeType.find_by(identifier: 'art', osm_value: 'art', icon: 'art_museum.png')
+    unless node_type.nil? && node_type.icon == 'art_museum.png'
+      node_type.icon = 'museum_art.png'
+      node_type.save
+    else
+      return
```
schultyy: `else return` is not necessary since there's no other statement after the whole conditional

db/migrate/20161128102235_change_art_museum_to_museum_art.rb
```diff
@@ -0,0 +1,27 @@
...
+    end
+  end
+
+  def down
+    node_type = LocalNodeType.find_by(identifier: 'art', osm_value: 'art', icon: 'museum_art.png')
+    unless node_type.nil? && node_type.icon == 'museum_art.png'
+      node_type.icon = 'art_museum.png'
+      node_type.save
+    else
+      return
```
schultyy: same as above

PR #487, opened 2016-11-28
Introduce latest translations from transifex

1000miles:
This PR is additional to PR #475 (Import/transifex translations nov2016) and introduces more translations from Transifex that contain changes in: categories.yml, authentication.yml & i18n folder.

Related issue ticket: 
- Add/Update place types (key/value-pairs) = categories #54


1000miles:
Thanks for review! 🍨

Issue #370, opened 2016-08-30
Add user infos to bug reports (mailto:)


holgerd:
As a community supporter I want to have additional user infos when receiving a bug report via email, so I can help the user much faster.
- user clicks "report bug" (does not exist yet)
- new email opens (mailto:) 
- starting text: 'please describe info here but only in German or English please'
- the mail is also pre-filled with additional info:
  - operating system
  - browser version
  - is user logged in?
    - user name
    - OSM connected?
      - OSM-URL of user page
  - current zoom level
  - current lat/long of map
  - language
  - filter status of map (status or categories or WC enabled/disabled)
- user adds his/her problem description to this email and clicks send.

Hoverbear:
The user agent should be sufficient for `operating system, browser version` and is probably the only way to get it in-browser.

Xylakant:
passing all this information to the mail client may be difficult. information can only be passed via query string and not all mail clients accept message bodies. It's not a properly specified thing but more of an ad-hoc semi standard. If we want more structured issues we should probably have our own "report issue" form on the page. This could in the background create an issue on our private github tracker or send an email if we prefer.

holgerd:
@Hoverbear Agree. I think some of those things can be looked up on the client side and for others we need to look it up in the app (eg username, loggedin status..)

1000miles:
@holgerd 

cc @schultyy 

**Estimation to fix:**
- approx. 2,5 -3 days

**Explanation:**
- concept & discussions
- user form
- mailing alerts
- pre-filled fields
- validations
- test cases

Hoverbear:
Probably related to #349

schultyy:
@holgerd Does the wheelmap have the information `OSM-URL of user page`? I can't find anything right now.

holgerd:
@schultyy If we have the OSM username the page of interest is www.openstreetmap.org/user/USERNAME/history eg. www.openstreetmap.org/user/holgerd/history

holgerd:
Let me know if/when I can test this and what the URL will be.

schultyy:
@holger: We also need a new translation for the "Report bug" menu entry: 
![screen shot 2016-10-04 at 11 56 53](https://cloud.githubusercontent.com/assets/235808/19070043/be160d82-8a29-11e6-9851-eaf36300000f.png)

Right now it always says "report bug" whatever language is selected.

holgerd:
[...]

PR #488, opened 2016-11-29
Feature/community support form

schultyy:
PR for #370 

This Pull Request brings the initial implementation of the community support form. It ships with the following features:

- A new form where the user can explain the issue
- A name field
- An email field

Additionally we gather the following information:
- User's operating system
- User's Browser manufacturer
- User's Browser version
- User is logged in yes/no
- User's OSM username (in case they're logged in)
- Currently selected language

What is missing right now:

- A translation for the menu entry in the header
- ~~The latest design applied to the form~~ (We now use the "relaunch" design here as well)
- The following data is not gathered yet:
   - current zoom level
   - current lat/long of map
   - filter status of map (status or categories or WC enabled/disabled)


1000miles:
Thanks for the PR. I have tested the form and here are some notes:

### 1. E-mail validation:
Currently we only validate if the email includes a `@` but not if the email is valid in general. Also there are no red marked indications as error messages after submit fails.



Would using a separate email validator like this make sense?
```
class EmailValidator < ActiveModel::EachValidator
  def validate_each(object, attribute, value)
    unless value =~ /\A([^@\s]+)@((?:[-a-z0-9]+\.)+[a-z]{2,})\z/i
      object.errors[attribute] << (options[:message] || I18n.t('help.new.form.email_invalid'))
    end
  end
end
```


### 2. Improve/Remove sentences from the support form output

a) If I send a support request I mainly expect to get a response, so does the sentence `We treat your address confidentially. We promise not to send you spam!` make sense here? This looks like a standard output from active record, or?

b) My suggestion here `Please describe info either in German or English here.` instead  of `Please describe info here but only in German or English please*` since the meaning is the same. What do you think?

app/mailers/community_support_mailer.rb
```diff
@@ -0,0 +1,9 @@
+class CommunitySupportMailer < ActionMailer::Base
+  default from: "noreply@wheelmap.org"
+
+  def send_to_support_team(community_support_request)
+    @community_support_request = community_support_request
+
+    mail(:to => "bugs@wheelmap.org", :subject => "Bugreport von wheelmap.org")
```
1000miles: If we change the wording of `Report bug` to `Help` then we should also change to e.g. `Help request from wheelmap.org` (or similar) here?

app/models/community_support_request.rb
```diff
@@ -0,0 +1,34 @@
+class CommunitySupportRequest
+  include ActiveModel::Validations
+  attr_accessor :name, :email, :message, :osm_username, :is_logged_in
+
+  validates_presence_of :name, :email, :message
+  validates :email, format: { with: /@/ }
```
1000miles: Please see my comment to a separate email validator here: https://github.com/sozialhelden/wheelmap/pull/488#issuecomment-263847821.

spec/controllers/community_support_controller_spec.rb
```diff
@@ -0,0 +1,143 @@
...
+
+  describe "GET #new" do
+    it "returns http success" do
+      get :new
+      expect(response).to have_http_status(:success)
+    end
+  end
+
+  describe "POST #create" do
+    let(:subject) { "Bugreport von wheelmap.org" }
```
1000miles: Same as above. "If we change the wording of Report bug to Help then we should also change to e.g. Help request from wheelmap.org (or similar) here?"

1000miles:
Another idea:

Right now we don't give user a success info after click on submit button. My suggestion here would be to display a flash message after the form could be committed.

Example for controller:
```
 flash[:success] = "Thank you for your request. We'll get back to you as soon as possible."
```

Xylakant:
The expression `/\A([^@\s]+)@((?:[-a-z0-9]+\.)+[a-z]{2,})\z/i` does not match some valid email addresses. For example `example@com` (only a TLD) is a valid email address, some NICs actually use addresses of that format. The format for valid addresses is quite complex and a regular expression describing the full set of RFC-conforming addresses takes about 1 full page: http://www.ex-parrot.com/pdw/Mail-RFC822-Address.html

There's also very limited use in checking the syntactical format since an RFC-conforming address does not guarantee that:

* The address will actually be processed correctly by real-world systems (they're not all standard conforming)
* The domain exists
* The domain accepts email at all
* The domain accepts email from us
* The user (local part) exists
* The user accepts email at all
* The user accepts email from us
* The user mailbox is currently available (not full, ...)
* The user that the mailbox belongs to is actually the user that filled in the form (typo that results in a different, still valid address)

If we'd want to check that all conditions required for email delivery to a given person, we'd have to send them an email. So any check that goes beyond a cursory "ok, there's an @ and something before and after that" is pretty much useless and most likely will reject valid addresses.

1000miles:
Got it. Thanks, @Xylakant. So we will stick to only check if there is `@` in the email and anything beyond it is overengineered. :)

schultyy:
@1000miles 
- I removed the sentence "We treat your address confidentially. We promise not to send you spam!"
- I double checked concerning the email address validation: We already validate if there's an`@` in it

app/models/community_support_request.rb
schultyy: As discussed, we only validate for `@` here.

spec/controllers/community_support_controller_spec.rb
schultyy: Sure.

app/mailers/community_support_mailer.rb
```diff
@@ -0,0 +1,9 @@
+class CommunitySupportMailer < ActionMailer::Base
+  default from: "noreply@wheelmap.org"
+
+  def send_to_support_team(community_support_request)
+    @community_support_request = community_support_request
+
+    mail(:to => "bugs@wheelmap.org", :subject => "Help request from wheelmap.org")
```
1000miles: So, we need to change the sentence as @holgerd would like to use the term `problem` instead of `help`. Suggestion for the subject: `Problem report from wheelmap.org`.

spec/controllers/community_support_controller_spec.rb
```diff
@@ -0,0 +1,143 @@
...
+
+  describe "GET #new" do
+    it "returns http success" do
+      get :new
+      expect(response).to have_http_status(:success)
+    end
+  end
+
+  describe "POST #create" do
+    let(:subject) { "Help request from wheelmap.org" }
```
1000miles: Please change subject to:  `Problem report from wheelmap.org` (pls see above comment).

app/assets/stylesheets/relaunch/_support.scss
```diff
@@ -0,0 +1,15 @@
+
+.support-form {
+  margin-top: $headerHeight + 20px;
+
+  textarea {
+    width: 100%;
+    height: 200px;
+  }
+
+  .help-block {
```
1000miles: What about naming the class `.error-text` here?

schultyy: It's a predefined class coming from formtastic so we can't rename it.

1000miles: I see. Good to know. Thanks! :)

1000miles:
LGTM, thanks. :)

PR #489, opened 2016-11-29
Fix unless statement in art museum migration

1000miles:
Change guard after loading node_type because previous guard did not handle the nil case properly


schultyy:
lgtm

PR #494, opened 2016-12-01
Feature/more keys for community support form

1000miles:
This PR is additional to PR #488 & PR #490 and does following:
- implements flash message after form submit
- converts previously hard coded strings to locale keys

Related Github issue:
#370


1000miles:
Thanks for the review! :)

Issue #496, opened 2016-12-01
[Form] Community Support Form submission takes ca. 20 seconds before success


1000miles:
At the moment when we submit a problem report via the new community support form it takes about 20 seconds until the request is successfully committed. This indicates that the form submission depends on the success of the email sending (mailer) which should not be the case because the user has to wait a while which in the end is less user-friendly.

To fix this we need a way to enqueue the mails which can be done via upgrade from Rails 4.1 to Rails 4.2 or we need to configure a delayed job for the community support form. 

Related Github issue: #370

1000miles:
Fixed in PR #501  and deployed on staging today (2016-12-13).

1000miles:
**Update, 2016-12-13:**

### Comparison tests:

#### Locally in VM (vagrant):

`.deliver`method (immediately):
  - display of flash note: 20 secs after submission

after switch to `.delay` method (delayed mail delivery):
  - display of flash note: 5-6 secs after submission
---

#### On staging server:

`.deliver`method (immediately):
  - display of flash note: 1-2 secs after submission

after switch to `.delay` method (delayed mail delivery):
  - display of flash note: 1-2 secs after submission

1000miles:
Done and deployed to production with release 20161219152440. Closing.

PR #501, opened 2016-12-09
Feature/send delayed email to support team

1000miles:
At the moment we trigger to send an email to the support team after the community support form submission. To prevent other performances being blocked by this, we have to enqueque and deliver those mails at a later time. This PR uses the simplest way to deliver delayed emails working for versions before Rails 4.2.

Resolves Github issue:
#496

### DIFF
--- a/spec/controllers/community_support_controller_spec.rb
+++ b/spec/controllers/community_support_controller_spec.rb
@@ -27,6 +27,10 @@ RSpec.describe CommunitySupportController, type: :controller do
       Delayed::Worker.delay_jobs = false
     end
 
+    after do
+      Delayed::Worker.delay_jobs = true
+    end
+
     context "with valid form params and user not logged in" do
       before do
         @current_locale = I18n.locale
--- a/spec/controllers/community_support_controller_spec.rb
+++ b/spec/controllers/community_support_controller_spec.rb
@@ -316,10 +316,6 @@ RSpec.describe CommunitySupportController, type: :controller do
         params
       end
 
-      after do
-        I18n.locale = @current_locale
-      end
-
       it "enqueues an email to send later to the support team" do
         submit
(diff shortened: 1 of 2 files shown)


spec/controllers/community_support_controller_spec.rb
```diff
@@ -20,10 +20,13 @@
     let(:latitude)        { 52.50327542986572 }
     let(:longitude)       { 13.411503732204435 }
     let(:last_zoom_level) { 15 }
+    let(:params) { params = {:community_support_request => { name: user_name, email: "holger@example.com", message: message}}}
+    let(:submit) { post :create, params }
```
schultyy: I'm not sure about this one because the naming is a bit confusing. It "submits" data to the service but actually memorizes the result.
Also is there a specific reason to use the `let` helper for this one?

spec/controllers/community_support_controller_spec.rb
```diff
@@ -306,6 +309,19 @@
       end
     end
 
+    context 'with valid form params and delayed email delivery' do
+      before do
+        ActionMailer::Base.deliveries.clear
+        Delayed::Worker.delay_jobs = true
+        params
```
schultyy: What does this call do here?

spec/controllers/community_support_controller_spec.rb
```diff
@@ -21,16 +21,17 @@
     let(:longitude)       { 13.411503732204435 }
     let(:last_zoom_level) { 15 }
     let(:params) { params = {:community_support_request => { name: user_name, email: "holger@example.com", message: message}}}
```
schultyy: Can you remove the assignment here, so it looks like this:

```ruby
let(:params) { {:community_support_request => { name: user_name, email: "holger@example.com", message: message}}}
```

PR #504, opened 2016-12-13
Move all navigation code into try block

### LINKED ISSUE #483: Bug:  New place without the address and geolocation
I tried to create a new place using  https://wheelmap.org/en/nodes/new/name_category. 
I didn't filled address and gelocation.
I couldn't create the point and also I didn't get any error message. It was only loading. No result.

schultyy:
Resolves #483 
This prevents navigation forward when the validation failed. Before the user didn't have the chance to correct what was wrong and where immediately taken on the next page.

/cc @lennerd


lennerd:
Good catch. I don't have a working setup right now, but the changes LGTM.

Issue #92, opened 2016-01-11
[LAYOUT] First list-item in NAV-bar shifted on mouse-over (in FF Browser)


1000miles:
In Chrome and Safari, it works great.
In Firefox 43.0.4 it's shifted. 

![00000030](https://cloud.githubusercontent.com/assets/4077916/12240307/f0f8d848-b88c-11e5-90b6-9b0b59244866.png)

1000miles:
Solved in PR #511. Closed.

PR #511, opened 2016-12-19
Fix padding & border in nav on mouse-over (Firefox)

1000miles:
Solves Github issue #92

Also solves part of https://github.com/sozialhelden/wheelmap/issues/370#issuecomment-267073839


1000miles:
As discussed with @schultyy this PR is approved.

Issue #519, opened 2016-12-21
[Locale] Switching from regional lang to another locale leads to `404 - not found`


1000miles:
The locale switcher fails when switching from a regional language to another locale because it tries to map a wrong url pattern.

- I go to wheelmap.org which has  german (de) as default language
- I switch from german (de) to português do brasil (pt_BR) => switch works fine! (https://wheelmap.org/pt_BR/map#/?zoom=10)
- When I try to switch from `pt_BR` to any other language (e.g. `es`) => I get a `404 error` because:
  - URL tries to map https://wheelmap.org/es/pt_BR/map#/?zoom=10 instead of https://wheelmap.org/es/map#/?zoom=10

![00000220](https://cloud.githubusercontent.com/assets/4077916/21399302/5615c2d0-c7ab-11e6-8329-f73bc62fa4d0.png)

- this seems to effect also other languages that are regional settings, e.g. `zh_TW`
- tested on production and staging with the same results

1000miles:
**Update, 2017-01-09:**

- deployed to staging, 2017-01-06
- deployed to production, 2017-01-09 with [release 20170109110124](https://github.com/sozialhelden/wheelmap/releases/tag/production-20170109110124)

PR #520, opened 2016-12-22
Fix/adapt regex for regional locales

1000miles:
The helper method [url_for_locale](https://github.com/sozialhelden/wheelmap/blob/master/app/helpers/application_helper.rb#L63-L78) tries to catch `pt-BR` while `pt_BR` is being used in wheelmap as regional locale pattern. Means when trying to switch e.g. from `/pt_BR/map` to Spanish the path would be `es/pt_BR/map` which causes `404 - not found`.

This PR corrects that pattern and adds some tests for missing locales.

Fixes Github issue: #519


1000miles:
Merci for the PR. LGTM.

PR #530, opened 2017-01-05
Fix toilet unknown bug

### LINKED ISSUE #492: bug: "toilets:wheelchair=unknown" written to OSM
- I am logged in as holgerd and use https://wheelmap.org/en/nodes/new/ to add the new node "Labyrintoom" to Wheelmap/OSM with wheelchair=no and no more info to the toilet, leaving it at "unknown" like this:

image

- I expect that only wheelchair=no is written to the node in OSM
- Instead, I also see toilets:wheelchair=unknown in the node: http://www.openstreetmap.org/node/4531573193 .
image

 I also see that there are thousands of tags like this in OSM: https://taginfo.openstreetmap.org/tags/toilets%3Awheelchair=unknown#overview (maybe all added this way?)

schultyy:
Resolves #492 

This prevents that unknown toilet status is being transmitted to Openstreetmap. This is already in place for the wheelchair status.

### DIFF
--- a/app/jobs/create_node_job.rb
+++ b/app/jobs/create_node_job.rb
@@ -1,7 +1,7 @@
 require 'htmlentities'
 class CreateNodeJob < Struct.new(:lat, :lon, :tags, :user, :client, :source)
-  WHEELCHAIR_TAG_KEY = 'wheelchair'
-  WHEELCHAIR_TOILET_TAG_KEY = 'toilets:wheelchair'
+  WHEELCHAIR_TAG_KEY = "wheelchair"
+  WHEELCHAIR_TOILET_TAG_KEY = "toilets:wheelchair"
 
   def self.enqueue(lat, lon, tags, user, source)
     raise "user not app authorized" unless user.app_authorized? # implies user.access_token.present?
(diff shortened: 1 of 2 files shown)


app/jobs/create_node_job.rb
```diff
@@ -8,7 +10,8 @@ def self.enqueue(lat, lon, tags, user, source)
     return unless Rails.env.production? || Rails.env.test?
 
     # Remove wheelchair tag if value is "unknown"
-    tags.delete("wheelchair") if tags["wheelchair"] == 'unknown'
+    tags.delete(WHEELCHAIR_TAG_KEY) if tags[WHEELCHAIR_TAG_KEY] == 'unknown'
+    tags.delete(WHEELCHAIR_TOILET_TAG_KEY) if tags[WHEELCHAIR_TOILET_TAG_KEY] == "unknown"
```
Hoverbear: Please use consistent quotes for both. :)

Hoverbear: Please use consistent quotes for both. :)

spec/jobs/create_node_job_spec.rb
```diff
@@ -85,7 +85,7 @@
   context "unknown value" do
 
     subject {
-      CreateNodeJob.enqueue(52.4, 13.0, { 'wheelchair' => 'unknown', 'amenity' => 'bar', 'name' => 'White horse' }, user, 'create_iphone')
+      CreateNodeJob.enqueue(52.4, 13.0, { 'wheelchair' => 'unknown', 'toilets:wheelchair' => 'unknown', 'amenity' => 'bar', 'name' => 'White horse' }, user, 'create_iphone')
     }
 
     it "does not save wheelchair tag" do
```
1000miles: Since we are using only one test case here, can you please adjust the test case description which also includes `wheelchair:toilet tag`?

Issue #526, opened 2017-01-04
[Menu] Remove nav item "Über wheelmap"


1000miles:
As agreed with @svenyo today we want to get rid of the nav item `Über wheelmap` in the main menu because of following reasons:
- we don't have enough space for the main menu anymore when adding `FAQ` and `Reiseführer` as new nav items
- `Über wheelmap` and `FAQ` would link to the same page (https://news.wheelmap.org/faq)

1000miles:
@svenja cc @holgerd 

If we remove the nav item `Über wheelmap` should we also remove the key (:about_wheelmap) from all `locale/relaunch.yml` or is it enough to only update the `en/relaunch.yml` file and send it to you for transifex?

Svenyo:
It should be enough to change it in the english source file. For the new "travel guide" in the menu we'd need translations anyways.

1000miles:
@Svenyo Merci for the quick feedback.

1000miles:
**Update, 2017-01-09:**

- deployed to staging, 2017-01-06
- deployed to production, 2017-01-09 with [release 20170109110124](https://github.com/sozialhelden/wheelmap/releases/tag/production-20170109110124)

PR #533, opened 2017-01-09
Update/main nav styling

1000miles:
This PR moves back to previous styling values to keep the main menu centered. Further infos can be found here: https://github.com/sozialhelden/wheelmap/issues/526.


1000miles:
Thanks, @hoverbear! 💃

PR #541, opened 2017-01-10
Integrate with make_osc & deploy on osm-database

Hoverbear:
There was an `make_osc.rb` and `string_hstore.rb` file on the deployed instances which was not in any of our repos. These are currently required for syncing between PostgreSQL and MySQL.

This PR converts configures things to appropriately talk to it.

### DIFF
--- a/config/deploy.rb
+++ b/config/deploy.rb
@@ -25,7 +25,7 @@ set :pty, true
 # Default value for :linked_files is []
 set :linked_files, %w{ config/database.yml config/open_street_map.yml config/metrics.yml config/librato.yml config/newrelic.yml .env config/secrets.yml}
 
-set :bundle_roles, [:app, :worker]
+set :bundle_roles, [:app, :worker, :importer]
 set :bundle_without, %w{ development test metrics deployment }.join(' ')
 set :bundle_jobs, 4
 set :bundle_binstubs, -> { shared_path.join('bin') }
@@ -75,7 +75,7 @@ namespace :deploy do
 
         # We need to link the existing shared `/var/apps/wheelmap/public/system` folder
         # into the new release.
-        execute :ln, "-s", "#{deploy_path}/public/system", "#{release_path}/public/"
+        execute :ln, "-s", "#{deploy_path}/public/system/uploads", "#{release_path}/public/system"
 
         # We create this file so the consul health check will pass. We can't use an
         # existing file since they are all unpredictably named.
--- a/config/deploy/production.rb
+++ b/config/deploy/production.rb
@@ -30,6 +30,7 @@ server 'app0.node.production', user: 'deploy', roles: %w{app}, port: 22
 server 'app1.node.production', user: 'deploy', roles: %w{app}, port: 22
 server 'asset.node.production', user: 'deploy', roles: %w{asset}, port: 22, :no_release => true
 server 'worker0.node.production', user: 'deploy', roles: %w{worker}, port: 22
+server 'osm-database.node.production', user: 'deploy', roles: %w{importer}, port: 22 # Used for replication.
 #server 'mysql.node.production', user: 'wheelmap', roles: %w{mysql}, port: 22
 
 # Custom SSH Options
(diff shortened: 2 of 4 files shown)


lib/tasks/make_osc.rake
```diff
@@ -0,0 +1,193 @@
+require 'pg'
+require 'builder'
+
+$fields = [ 'shop', 'office', 'aerialway', 'aeroway', 'amenity', 'tourism', 'historic', 'sport', 'leisure', 'public_transport' ]
```
schultyy: Is there a reason this has to be a global variable? I'm asking because of the `$`.

lib/tasks/make_osc.rake
```diff
@@ -0,0 +1,193 @@
...
+
+# encoding: UTF-8
+namespace :make_osc do
+  desc 'Make OSC files.'
+  task :full => :environment do
+    $conn = PGconn.open(:dbname => 'osm', :user => 'osm', :password => 'osm', :host => 'osm-database')
```
schultyy: Same as above? Does this one have to be global?

lib/tasks/make_osc.rake
```diff
@@ -0,0 +1,193 @@
...
+                    end
+                end while true
+            end
+        end
+    end
+    FILE_HANDLE.close
+  end
+
+  task :diff => :environment do
+    $conn = PGconn.open(:dbname => 'osm', :user => 'osm', :password => 'osm', :host => 'osm-database')
```
schultyy: Same as above with `$conn`

Gemfile
```diff
@@ -3,6 +3,7 @@ source 'https://rubygems.org'
 group :default do
   gem 'rails', '~> 4.1.0'
   gem 'mysql2', '~> 0.3.18'
+  gem 'pg', '~> 0.18.4'
```
schultyy: The `pg` gem requires some native extensions to be build on installation. For that it requires the Postgres dev packages. By having it in the `:default` group we require all devs to have Postgres installed on their machines when they want to work on wheelmap.

I'd propose to have a separate `group` for this we then can exclude during installation via `bundle install --without`.

lib/tasks/make_osc.rake
Hoverbear: I don't know.

lib/tasks/make_osc.rake
Hoverbear: I don't know. I didn't write this code.

lib/tasks/make_osc.rake
Hoverbear: I don't know. I didn't write this code.

Gemfile
Hoverbear: Hm, this seems reasonable.

Hoverbear:
We discussed removing the `$` from the variables. Will follow up with this.

Hoverbear:
@schultyy Those should be resolved.

Hoverbear:
Requested changes made and work.

Xylakant:
I would consider stopping this effort: What's the gain compared to the effort invested and the dependencies introduces into the main app. I'd rather see this in a small standalone script with a gemfile that's separate from the main wheelmap app.

Hoverbear:
@Xylakant So a separate repo which we also deploy via capistrano or?

Xylakant:
I'm uncertain if cap is not an overkill in that case. This will not see many updates, installing that on the machine via chef should be perfectly fine.

Hoverbear:
@Xylakant Well, the only dependency we introduce is `pg` which we've moved into the `production` group anyways so it can be excluded. It's a rather important part of a deployment of wheelmap anyways and I'm concerned placing it *somewhere else* would cause it to get forgotten/lost (again).

Xylakant:
The use of a group is somewhat limited, people expect `bundle install` to install all development dependencies and excludes should be limited to environments that are production or close to. It also requires us to install pg on all app machines that should not require it currently.

I agree that it should be committed somewhere - it can even live in a "tools" or "osm-sync" subfolder of this repo, but I don't think the effort of folding this into the rake task has any positive ROI either in time or reduced complexity.

Hoverbear:
So my reasoning for making it a rake task was that it was called by a rake task already, so it seemed easy to integrate that way.

I suppose it wouldn't be much harder to drop this in a subfolder and give it it's own gemfile then modify capistrano to have the dependencies installed when it is deployed.

Xylakant:
as I said - I don't think that installing it via CAP has any positive ROI either, but that's a minor point. I'd really go and move it to a dedicated repo and treat it like we treat any other dependency: We assume that it's available at the point where we install the app.

Hoverbear:
Ok.

Hoverbear:
Okay, this should be good to go.

Hoverbear:
Rebased on top of `feature/infra-rebuild`

Hoverbear:
Rebased.

Issue #538, opened 2017-01-10
[FE/Design] Locale dropdown not rendering consistently on all pages


1000miles:
**Background:**
- I go to `wheelmap.org` (default language: DE)
- I click on the nav item `Problem melden`
- on https://wheelmap.org/community_support/new I expect to see the locale switcher dropdown to be rendered properly on all pages instead I see a not fully rendered locale switcher dropdown

**Important to know:**
- this behavior does not occur consistently, means on some pages it renders the dropdown, icon & font-size correctly
- I am not sure if there is already a ticket (I have searched for it but couldn't find) but I know that this issue exists since a while. With 2 less nav items in the past we could ignore but now it's not possible anymore. 

#### on `Problem melden` page:

<img width="1268" alt="00000235" src="https://cloud.githubusercontent.com/assets/4077916/21803783/f6b0d9f2-d72b-11e6-9fd1-38fd2b6c4bb4.png">

####  on `Problem melden` page using `DE` as language:

<img width="1271" alt="00000236" src="https://cloud.githubusercontent.com/assets/4077916/21803954/d5604b10-d72c-11e6-86ae-50a3e988d236.png">

----

#### on root page using `DE` as language:

<img width="1265" alt="00000237" src="https://cloud.githubusercontent.com/assets/4077916/21804853/6977d8b4-d731-11e6-97be-69716fb11ff6.png">

#### on root page using `pt_BR` as language:

<img width="1272" alt="00000238" src="https://cloud.githubusercontent.com/assets/4077916/21804985/29b9a9a4-d732-11e6-8ed5-b0d578e87a07.png">

1000miles:
@Hoverbear Unfortunately this ticket is not fixed yet. Locally your fix (PR #544) seems to have fixed it but when deploying to staging and production the issue is still existent. Can you please have a further look into this? Merci. :)

<img width="1270" alt="00000259" src="https://cloud.githubusercontent.com/assets/4077916/22884658/5f402fec-f1f6-11e6-802a-d1aa104bffb4.png">

1000miles:
As agreed with @holgerd today I am assigning this ticket temporarily to Holger.

holgerd:
@opyh can you maybe help out here?

Hoverbear:
#581 resolves this. Was due to some misconceptions on my part regarding how rails works.

schultyy:
We today merged and deployed Andrew's pull request containing the bugfix. I therefore close this ticket now.

PR #544, opened 2017-01-12
Fix language picker not being styled on some pages

Hoverbear:
Contributes to fixing #538.

This may not have caught everything.

### DIFF
--- a/app/views/community_support/new.html.haml
+++ b/app/views/community_support/new.html.haml
@@ -1,4 +1,4 @@
-.container-fluid
+.container-fluid#wheelmap
   .support-form
     %h1=t('community_support_form.new.form.title')
     %p=t('community_support_form.new.form.header_text_html')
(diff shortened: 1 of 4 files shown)


app/views/community_support/new.html.haml
```diff
@@ -12,3 +12,10 @@
       = f.input :email, hint: ""
       = f.input :message, as: :text , label: I18n.t('community_support_form.new.form.describe')
       = f.action :submit, as: :button, label: I18n.t('community_support_form.new.form.submit_button')
+
+- content_for :javascript do
+  = javascript_include_tag "jquery"
+  = javascript_include_tag "bootstrap"
+  = javascript_include_tag "bootstrap-select"
+  %script
+    $(document).ready(function () { $('.selectpicker').selectpicker(); });
```
1000miles: @Hoverbear Thanks for the PR. 

Does it make sense to use a partial instead of repeating the code blocks in 3 different files? 

Example:
```
render :partial => 'shared/available_locales'
```

Source: [available_locales](app/views/shared/_available_locales.html.haml)

Hoverbear: I thought about this as well. I didn't really have an opinion. What do you think?

1000miles: i would go for less repeating code. :)

1000miles: @hoverbear will you update this PR or shall i take over this ticket? :)

Hoverbear: I will get it, sorry! :) Will be done tomorrow end of day at latest.

1000miles: Merci. :)

PR #549, opened 2017-01-16
Add places nearby from accessibility.cloud

opyh:
This adds places nearby from partners that are registered on [accessibility.cloud](https://www.accessibility.cloud/).

The token is registered for Wheelmap as app of Sozialhelden. @holgerd has admin access for the app’s page on accessibility.cloud.

You can test this with locations that have other locations nearby on accessibility.cloud, e.g. `/nodes/3291269846` (a place in Vienna where accessibility.cloud has a listing of public toilets nearby) of `/nodes/3584462192` (for this place, other PoIs from [AXS Map](http://www.axsmap.com/) are listed).

screen shot 2017-01-16 at 16 06 29


opyh:
The categories are currently shown in English, but this is being fixed on the server side soon, so no change to the PR necessary.

Hoverbear:
Seems to sometimes place an `[object Object]`, is this expected?

screen shot 2017-01-17 at 09 28 51

opyh:
Thanks for the report. Of course this is not expected :) I will have a look.

opyh:
@Hoverbear The bug you reported is fixed now. The widget doesn't contain any [object Object] translations anymore. With the new commits, it is only displayed if there are results.

vendor/assets/javascripts/accessibility.cloud.1.2.8.min.js
```diff
@@ -0,0 +1,20 @@
...${y}</span> <span class="numeric">${x}/5</span>`}function h(w){if($.isArray(w))return`<ul class="ac-list">${w.map(x=>`<li>${h(x)}</li>`).join('')}</ul>`;if($.isPlainObject(w)){const x=$.map(w,(y,z)=>{if(z.match(/Localized/))return'';const A=d(z,w);return $.isArray(y)||$.isPlainObject(y)?'areas'===z&&1===y.length?h(y[0]):`<li class="ac-group"><header class='subtle'>${A}</header> ${h(y)}</li>`:z.startsWith('rating')?`<li class="ac-rating">${A}: ${g(parseFloat(y))}</li>`:`<li class="ac-${typeof y}">${A}: <span class='value'>${f(y)}</span></li>`});return`<ul class="ac-group">${x.join('')}</ul>`}return w}function k(w){return w.accessibility&&w.accessibility.accessibleWith&&w.accessibility.accessibleWith.wheelchair}var l=c(6),m=c.n(l),n=c(5),o=c.n(n),p=c(3);c.n(p);var q=c(1);Object.defineProperty(b,'__esModule',{value:!0});const u=q.a,v={getLocale(){return void 0||this.locale},apiDomain:'https://www.accessibility.cloud',getPlacesAround(w){return $.ajax({dataType:'json',url:`${this.apiDomain}/place-infos?includeRelated=source&locale=${this.getLocale()}`,data:w,headers:{Accept:'application/json','X-Token':this.token}})},resultsTemplate(){return`<ul class="ac-result-list" role="treegrid"> \
```
Hoverbear: Perhaps Accessibility Cloud could host this library at some point in the future. For now this is fine.

app/views/nodes/_node_nearby.html.haml
```diff
@@ -0,0 +1,26 @@
+%section.node-nearby-locations-from-partners
+  %h2=t('.places-nearby')
+  %section.node-nearby-locations-from-partners-list
+
+- content_for :javascript do
+  :javascript
+    window.addEventListener("load", function load(event) {
```
Hoverbear: According to MDN the [`DOMContentLoaded`](https://developer.mozilla.org/en/docs/Web/Events/DOMContentLoaded) should probably be used in this situation instead of `load`.

>  It is an incredibly popular mistake to use load where DOMContentLoaded would be much more appropriate, so be cautious.

Using this you won't need to have the `window.removeEventListener` line. The final argument to this function would also be unnecessary I think.

opyh:
@Hoverbear Wow, thanks for all that feedback, this is really valuable. Luckily wheelmap is the first website to use the widget, so it's no big deal to change the code according to your suggestions (all of them).

As the changes would not touch the other infrastructure of wheelmap, would it be okay if you merge this and I make a second pull request next week so we can show the feature earlier?

Hoverbear:
@opyh Great. :) Yes we can merge this *almost* as is, we noted yesterday deployment isn't very friendly to the minified javascript file you included in the vendor. Please use the non-minified and I can merge this!

opyh:
Ah, I was using https://github.com/babel/babili for minification, but probably it’s too early to assume usual toolchains accept JS libraries with ES2015+ things… Good to know. Using non-minified version now!

Hoverbear:
Great. :)

app/assets/javascripts/nodes.coffee
```diff
@@ -1,6 +1,5 @@
 #= require jquery
 #= require jquery_ujs
-#= require accessibility.cloud.1.2.8.min.js
```
Hoverbear: So now it's not imported at all and the functionality doesn't work. :(

opyh:
@Hoverbear Whew, how could I not see that! I thought I had tested the page locally again but seems I did miss something. Fixed now.

opyh:
Yep, exactly. Looking forward to seeing this live…

schultyy:
Deploying this to staging right now

schultyy:
Deployed on Staging.

schultyy:
Is deployed to production now.

app/views/nodes/_node_nearby.html.haml
opyh: Yep, this will make the loading a bit faster too. Will do it.

opyh: (It was `load` because it's taken from the example script — there, you can't know if the JS is already loaded. I considered using a callback pattern but that's more error prone for beginners. In your situation the JS is put through the asset pipeline so `DOMContentLoaded` makes most sense)

app/views/nodes/_node_nearby.html.haml
```diff
@@ -0,0 +1,26 @@
...
+      window.removeEventListener("load", load, false); // remove listener, no longer needed
+      AccessibilityCloud = window['accessibility.cloud'].default;
```
opyh: `window['accessibility.cloud'].default` is not beautiful, you're right. It's just a workaround to access the module without a modular JS build system. The applicable namespace here is not browser JavaScript, but npm.

AccessibilityCloud is going to be a constructor soon, so let's keep the naming :)

Btw, the Mozilla guidelines don't state that **only** constructors are meant to have an uppercase first letter. Libraries that I know usually have this case for namespaces/classes that only have static properties.

And with React/ES6/webpack/npm, you'll be able to use `import AccessibilityCloud from 'accessibility.cloud';`, which is more elegant. Even more elegant would be to have a React component then, but that's for a bit later.

app/views/nodes/_node_nearby.html.haml
```diff
@@ -0,0 +1,26 @@
...
+      AccessibilityCloud.token = '7f039b60e27a4d02b13c5ad79fbe9d7b';
+      AccessibilityCloud.locale = I18n.locale;
```
opyh: Yes, good idea, using theses as parameters to check is better, uses less state, and is more functional-style. Will do that.

app/views/nodes/_node_nearby.html.haml
```diff
@@ -0,0 +1,26 @@
...
+      var containerElement = document.querySelector('section.node-nearby-locations-from-partners');
+      var element = document.querySelector('section.node-nearby-locations-from-partners-list');
+      var parameters = {
+        latitude: #{@node.lat},
+        longitude: #{@node.lon},
+        accuracy: 1000,
+        limit: 20,
+        excludeSourceIds: 'LiBTS67TjmBcXdEmX', // excludes Wheelmap's own PoIs from results
```
opyh: Hmm, I'd love to have friendly names for sources! I'll check how we can do that (mid-term). It's more work than with (for example) Rails, where the community took care of this already. I've got to much other things to write my own lib for this, and there are several 3rd party libs for our stack, but they come with a lot of monkey patching so I'm a bit wary ;)

app/views/nodes/_node_nearby.html.haml
```diff
@@ -0,0 +1,26 @@
...
+      };
+      AccessibilityCloud.loadAndRenderPlaces(element, parameters).done((response) => {
```
opyh: Arrow function: Yes, sorry, already writing too much ES6. Having our own documented callbacks makes sense, too.

app/views/nodes/_node_nearby.html.haml
```diff
@@ -0,0 +1,26 @@
...
+        if (response && response.features && response.features.length) {
+          containerElement.className = containerElement.className += ' has-results';
```
opyh: Absolutely.

app/views/nodes/_node_nearby.html.haml
opyh: For the `excludeSourceIds` parameter, an arrays work too. Will have to document all this too.

Hoverbear: It might be worthwhile just making it always an array. It's really easy to accidentally use a string as an array in JS and this will likely make code simpler. :)

app/views/nodes/_node_nearby.html.haml
Hoverbear: I totally relate. :)

PR #557, opened 2017-01-19
Search string stays consistent

### LINKED ISSUE #97: Search not user friendly: spaces are replaced with "+" when re-searching, leading to 0 results
When I search for "12 place voltaire 94200" I get one result. I select the result and the map shows that area.
After that I hit the search button again but then I get no results at all, because the search term was changed to `12+place+voltaire+94200`. From a user's perspective this makes the search functionality difficult to use.

Hoverbear:
Should fix #97.


1000miles:
Thanks for the PR. Approved! :)

 🙅🏽

PR #559, opened 2017-01-20
Add mapillary to node detail view

### LINKED ISSUE #480: Add Mapillary to detail pages
Suggestion from the OSM forum:
Add Mapillary and OpenStreetView to detail pages
https://forum.openstreetmap.org/viewtopic.php?id=55958

Sounds like a good idea, maybe as an addition instead of a replacement of Google Streetview.

https://github.com/mapillary/mapillary-js seems to be free for non-profits and I did not find a widget option for OSV yet.

schultyy:
Resolves #480 

This Pull Request adds Mapillary Street View for the Node Detail View additionally to Google Street View.
It is shown only if there's content available for the given latitude and longitude, otherwise it stays hidden.


config/environments/staging.rb
```diff
@@ -6,7 +6,7 @@
   config.cache_classes = true
 
   # Full error reports are disabled and caching is turned on
-  config.consider_all_requests_local       = false
+  config.consider_all_requests_local       = true
```
Hoverbear: Why is this done?

schultyy: Ah good catch! I only enabled that to debug some things. This should be set to `false` obviously.

PR #565, opened 2017-02-01
Update accessibility.cloud JS widget to 2.2.0

opyh:
- Show distances in miles/yards if locale is set to English
- Distances are now links that open OpenStreetMaps, fixing usability of displayed PoIs that do not have their own website
- Update accessibility.cloud access token
- Improve AccessibilityCloud API as discussed in #549
- Convert the JS widget’s translation key string to underscore_case


opyh:
Don't worry about the minified version being included again – this time it's minified with Uglify instead of Babili so all should be well with the asset pipeline now :)

app/views/nodes/_node_nearby.html.haml
```diff
@@ -1,26 +1,28 @@
...
-  %h2=t('.places-nearby')
+  %h2=t('.places_nearby')
   %section.node-nearby-locations-from-partners-list
 
 - content_for :javascript do
   :javascript
-    window.addEventListener("load", function load(event) {
-      window.removeEventListener("load", load, false); // remove listener, no longer needed
+    window.addEventListener("DOMContentLoaded", function load(event) {
+      window.removeEventListener("DOMContentLoaded", load, false); // remove listener, no longer needed
```
opyh: Alright, removed it :)

app/views/nodes/_node_nearby.html.haml
```diff
@@ -1,26 +1,28 @@
...
-        excludeSourceIds: 'LiBTS67TjmBcXdEmX', // excludes Wheelmap's own PoIs from results
+        accuracy: 1000, // include places within 1 km radius
+        limit: 10,
+        excludeSourceIds: ['LiBTS67TjmBcXdEmX'], // excludes Wheelmap's own PoIs from results
       };
-      AccessibilityCloud.loadAndRenderPlaces(element, parameters).done((response) => {
-        if (response && response.features && response.features.length) {
-          containerElement.className = containerElement.className += ' has-results';
+      accessibilityCloud.loadAndRenderPlaces(element, parameters, function(error, response) {
+        if (!error && response && response.features && response.features.length) {
```
opyh: 1) Possible errors are broken responses and HTTP errors. Both cases fail silently here, the widget is simply not shown then. Users don't need to see them as the feature is just an addition, not core functionality of the page, but it would be nice if developers could see them, so if you have a solution for client-side error logging or monitoring, I'd include it here! How would I do this? Or could you even do it? ;)

2) The code is to ensure the widget's title is not shown when the list is empty (a successful response can be empty). The parent attribute extra check is for the case that the data structure is not as expected (should not happen, but you never know…), in this case I want to avoid an exception. The error parameter already carries the information if the response was successful or not.

schultyy: 1. A silent fail is I think Ok in this case because the user couldn't do something about it anyways. I think concerning client-side error logging `console.log` should be sufficient because we don't have anything in place to collect logs from clients.
2. Ahh I see. that makes sense.

opyh:
The code clearly differentiates error handling and empty result lists.

schultyy:
Is now available on Staging.

Issue #558, opened 2017-01-20
I'd like to get nodes from the API that are changed since a given timestamp


opyh:
Hi,

I'm Sebastian Felix Zappe, currently working on accessibility.cloud 👋

1) For the project it would be great to have a way to get nodes from the API that have recently been changed so accessibility.cloud could fetch nodes without downloading a big dataset when syncing, saving bandwidth, CPU and our time. Is this already possible? If not, how could we make this possible?
2) To really have a data synchronization, an endpoint to get IDs of deleted nodes since a specified timestamp would be helpful.

@holgerd Could you see if this is possible and prioritize? :)

schultyy:
So when can provide changed nodes easily already, for deleted ones we need to start writing some sort of log (another table just containing the Ids).

holgerd:
yes

schultyy:
@opyh I'm currently thinking of the frequency you might query this data. Is there a need to have them immediately after something changed or do you fetch them in longer intervals?

schultyy:
We definitely have to change the import tasks for this, because they're the main actor when it comes to delete data. 
We probably also have to have a look at ActiveAdmin as well, because via the Admin interface a user can delete a POI as well.

Xylakant:
using a trigger in the database may be an option, but much less visible.

schultyy:
Maybe it makes sense to copy the CouchDB changes stream behavior.

schultyy:
@opyh Thuy and me are currently talking about how the API response should look like. A first draft looks like this:

```json
[
  {
    "osm_id": 1,
    "action": "update",
    "updated_at": "30.01.2017 9:30"
  },
  {
    "osm_id": 2,
    "action": "delete",
    "updated_at": "29.01.2017 9:30"
  }
]
```

Does this already cover your requirements or do we have to include more information such as which field(s) have changed?

opyh:
Cool, thank you very much, this should work! :) The server will just get the whole PoI again on re-import, no matter which fields changed exactly, so nothing else needed from AC's side.

opyh:
If I can have one wish: Can we se UTC timestamps in milliseconds instead of date strings?

Xylakant:
[...]

PR #567, opened 2017-02-07
558 feat/node changes stream

1000miles:
This PR adds a new API endpoint for node changes, updated and deleted nodes. The client can specify the start date to search for. 

**Based on following cases:**
  - return a response if the poi has been changed
  - return a response if the poi has been deleted (not existent)
  - return  an empty response if  the date is in the future (not existent)
  - return an error response when no date has been submitted

**we implemented the task as follows:**
- we added another table in the database to log via callbacks as soon as a poi has been changed or deleted

Resolves Github ticket #558.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -117,6 +117,7 @@ group :test do
   gem 'shoulda-matchers', '~> 3.1'
   gem 'syntax'
   gem 'email_spec'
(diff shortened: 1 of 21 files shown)


1000miles:
Thank you! 🍰

PR #569, opened 2017-02-09
Forbid anonymous users to create and update nodes

schultyy:
Related to #477 

This PR is a first step in the direction of ensuring that `wheelmap_android` and `wheelmap_visitor` are not able to create or update nodes via the wheelmap API.
Their only use case is to ensure that anonymous users are able to change the wheelchair and toilet status of an existing POI without the need of having an OpenStreetMap account.

We realize this change by introducing a new flag for users called `anonymous`. This is set to `false` for all users except `wheelmap_android` and `wheelmap_visitor`. Before a `create` or `update` action is performed we check if the current user has this flag set. If yes, the request is denied with a 403. This will result in failing requests for very old Android deployments (installations from the time span between early 2011 and late 2012) and breaks with expected behavior. This is breaking change is intended because the business requirement clearly says that `wheelmap_android` cannot create/update POIs.

### DIFF
--- a/db/migrate/20170217131434_change_anonymous_flag_for_wheelmap_visitor_and_wheelmap_android.rb
+++ b/db/migrate/20170217131434_change_anonymous_flag_for_wheelmap_visitor_and_wheelmap_android.rb
@@ -1,0 +1,23 @@
+class ChangeAnonymousFlagForWheelmapVisitorAndWheelmapAndroid < ActiveRecord::Migration
+
+  EMAILS = [ 'visitor@wheelmap.org', 'android_app@wheelmap.org']
+
+  def up
+    change_anonymous_flag(true)
+  end
+
+  def down
+    change_anonymous_flag(false)
+  end
+
+  private
+  
+  def change_anonymous_flag(anonymous_flag)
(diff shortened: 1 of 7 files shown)


1000miles:
For unclear reasons the migration for the `tango measurement` has changed and failed when I ran `bundle exec rake spec` from my virtual machine even though this part shouldn't be involved here. This issue is fixed for now but we need to observe this behavior.

db/migrate/20170217131434_change_anonymous_flag_for_wheelmap_visitor_and_wheelmap_android.rb
```diff
@@ -0,0 +1,23 @@
+class ChangeAnonymousFlagForWheelmapVisitorAndWheelmapAndroid < ActiveRecord::Migration
```
Hoverbear: Hm... These two users are not created by any seeds or migrations. Running this migration on a new database does not actually make these two users anonymous.

Any newly created production environment using these two users will need to manually go and set these users to anonymous after creating them (also a manual process).

PR #570, opened 2017-02-10
Paginated changes stream

schultyy:
Follow up for #567 
Related to #558 

To make the usage of the node changes stream more convenient, this PR adds pagination to it. 

Example:

```bash
$ curl -XGET "http://localhost:3000/api/nodes/changes?api_key=hktndutjMMjTaxDzsBJz&since=2017-01-01&page=5&per_page=1" | underscore pretty

  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   222    0   222    0     0   6119      0 --:--:-- --:--:-- --:--:--  6166
{
  conditions: { page: 5, per_page: 1, format: "json", locale: "de" },
  meta: { page: 5, num_pages: 6, item_count_total: 6, item_count: 1 },
  pois: [
    {
      osm_id: 10672204,
      action: "update",
      timestamp: "2017-02-07T13:44:48.000Z"
    }
  ]
}
```

This adds `meta` and `conditions` to the json response containing information about the (paged) response.

Although the URL stays the same, the changes stream implementation moved into its own controller for the following reasons:

- The changes stream in its very first version only operated on `Pois` so
  it was obvious to put it in the `NodesController`
- But now all data is gathered from a separate model namely `PoiLog`
- We want to have pagination implemented in the same technical way as in all the other controllers
- The NodesController operates on `Pois` but we use `PoiLogs`


schultyy:
Deployed on Staging. @opyh can you test if it works as expected?

opyh:
Works fine for me, the importer uses the pagination and seems to work fine. Thank you very much!

Issue #577, opened 2017-02-20
Use Poi.destroy_all instead of Poi.delete_all in PlanetReader


schultyy:
We found out that our newly created `PoiLog` only contains updated entries but no deleted ones. 

The reason for this is the following:
We invoke `Poi.delete_all` in `PlanetReader` and `delete_all` does not invoke any callbacks. 

In order to fix this, we need to change `delete_all` to `destroy_all`. `destroy_all` invokes callbacks but also loads each entry from database first. This might result in decreased performance but is the most reasonable way of fixing it right now.

1000miles:
Deployed to [production](https://github.com/sozialhelden/wheelmap/releases/tag/production-20170222124705) today.

PR #579, opened 2017-02-22
Fix/change to destroy all in planet reader

1000miles:
This PR fixes the issue that currently only updated pois but no deleted pois are logged.

Related Github issue: https://github.com/sozialhelden/wheelmap/issues/577


schultyy:
The reason is that `delete_all` does not invoke any callbacks on the models but directly issues a `DELETE FROM pois WHERE osm_ids IN ()` statement. On the contrary `destroy_all` invokes all involved callbacks (which is crucial in our case) but has one tradeoff: It loads the involved records first to then issue `DELETE` statements for each of them.

PR #593, opened 2017-03-29
Move vagrantfile into correct location

Hoverbear:
I didn't notice the `Vagrantfile` was moved in https://github.com/sozialhelden/wheelmap/pull/431 after the PR approval when I merged. This moves it back to match documentation and fit into the typical use case.


emig:
👍 much better, less changes needed and ok, Vagrant file at root seems the standard.

PR #437, opened 2016-10-13
Migrate to mailgun

Hoverbear:
Migrates the application to use mailgun for mailing.

Addresses https://github.com/sozialhelden/wheelmap/issues/436 temporarily

Issue #349, opened 2016-08-01
Use mailgun for sending of transactional mails


holgerd:
To ensure emails are delivered from the app to our users, I want to use a dedicated service to ensure the mails are not marked as spam.

Thuy suggested mailgun and Holger has setup an account (see 1pw)

lennerd:
@schultyy @1000miles Do you already have some test rutines or something to test new email providers?

@holgerd I do not have access to 1pw yet.

holgerd:
this is done, right?

Hoverbear:
It's done on infra-rebuild, I will work on backporting this to current.

Hoverbear:
These are the changes: https://github.com/sozialhelden/wheelmap/pull/437/files

Some changes are needed in the secrets.yml as well.

PR #594, opened 2017-03-29
Set up configuration for mailgun

Hoverbear:
Based off https://github.com/sozialhelden/wheelmap/pull/437.
Contributes to https://github.com/sozialhelden/wheelmap/issues/349.

**This requires changes to `config/secrets.yml` before deployment.** The changes required can be found in the `infra-rebuild` branch of the chef repositories.

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -65,17 +65,18 @@ Wheelmap::Application.configure do
   # config.action_mailer.raise_delivery_errors = false
   config.action_mailer.raise_delivery_errors = true
   config.action_mailer.delivery_method = :smtp
-  config.action_mailer.default_url_options = { host: 'wheelmap.org' }
+  config.action_mailer.default_url_options = {
+    :host => 'wheelmap.org',
+    :from => 'noreply@wheelmap.org',
+  }
 
   config.action_mailer.smtp_settings = {
-    address: 'mail.wheelmap.org',
-    enable_starttls_auto: false,
-    port: 25,
-    domain: 'wheelmap.org',
-    authentication: :login,
-    user_name: Rails.application.secrets.mail_user,
-    password: Rails.application.secrets.mail_password
-
+    :authentication => :plain,
+    :address        => 'smtp.mailgun.org',
+    :port           => 587,
+    :domain         => 'mg.wheelmap.org',
+    :user_name      => Rails.application.secrets.mail_user,
+    :password       => Rails.application.secrets.mail_password
   }
 
   # Enable threaded mode
--- a/config/environments/staging.rb
+++ b/config/environments/staging.rb
@@ -62,17 +62,18 @@ Wheelmap::Application.configure do
   # config.action_mailer.raise_delivery_errors = false
   config.action_mailer.raise_delivery_errors = true
   config.action_mailer.delivery_method = :smtp
-  config.action_mailer.default_url_options = { host: 'staging.wheelmap.org' }
+  config.action_mailer.default_url_options = { 
+    :host => 'staging.wheelmap.org',
+    :from => 'noreply@wheelmap.org',
+  }
 
   config.action_mailer.smtp_settings = {
-    address: 'mail.wheelmap.org',
-    enable_starttls_auto: false,
-    port: 25,
-    domain: 'wheelmap.org',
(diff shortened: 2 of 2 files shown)


config/environments/production.rb
```diff
@@ -65,17 +65,16 @@
...
   config.action_mailer.smtp_settings = {
-    address: 'mail.wheelmap.org',
-    enable_starttls_auto: false,
-    port: 25,
-    domain: 'wheelmap.org',
-    authentication: :login,
-    user_name: Rails.application.secrets.mail_user,
-    password: Rails.application.secrets.mail_password
-
+    :authentication => :plain,
```
1000miles: So when we are using `plain` for authentication shouldn't we then set `enable_starttls_auto: true`?

Hoverbear: I based the configuration off https://devcenter.heroku.com/articles/mailgun#sending-emails-via-smtp ... You are probably right though!

Hoverbear: > `:enable_starttls_auto` - Detects if STARTTLS is enabled in your SMTP server and starts to use it. Defaults to `true`.

It defaults to `true`: http://api.rubyonrails.org/v5.1.0/classes/ActionMailer/Base.html

config/environments/production.rb
```diff
@@ -65,17 +65,16 @@
...
+    :address        => 'smtp.mailgun.org',
+    :port           => 587,
+    :domain         => 'mg.wheelmap.org',
```
1000miles: Why aren't we use env variables for `user_name` and `password` here?

config/environments/staging.rb
```diff
@@ -62,17 +62,16 @@
...
-    port: 25,
-    domain: 'wheelmap.org',
-    authentication: :login,
-    user_name: Rails.application.secrets.mail_user,
-    password: Rails.application.secrets.mail_password
-
+    :authentication => :plain,
+    :address        => 'smtp.mailgun.org',
+    :port           => 587,
+    :domain         => 'mg.wheelmap.org',
```
1000miles: Same here. Why aren't we use env variables for `user_name` and `password` here?

config/environments/production.rb
1000miles: True. :)

config/environments/production.rb
Hoverbear: We don't currently do it now. That would be a separate PR, perhaps?

1000miles: Why not in this PR? :)

Hoverbear: Because `config/secrets.yml` seems like an appropriate place to do it.

1000miles: Right, but how do you want to fetch the credentials after you put them into the secrets.yml file?

From what I understand so far we  basically put in `config/secrets.yml`  the credentials:

```ruby

# Production Mailgun

production:
  mail_username: <%= ENV["mail_username"] %>
  mail_password: <%= ENV["mail_password"] %>
```
And call it e.g. from `config/environment/production.rb`:

```ruby
  config.action_mailer.smtp_settings = {
    :user => ENV["mail_username"],
    :password => ENV["mail_password"],
    :authentication => :plain,
    :address        => 'smtp.mailgun.org',
    :port           => 587,
    :domain         => 'mg.wheelmap.org',
}
```

See also: https://github.com/rails/rails/tree/4-1-stable/actionmailer

1000miles: oops, i accidently put ruby embedded code styling (ERB), should be this as an example in `secrets.yml`:

```ruby

# Mailgun

production:
  mail_username: "mail_username"
  mail_password: "mail_password"

```

Hoverbear: Ah, right. I missed these lines somehow: https://github.com/sozialhelden/wheelmap/pull/437/files#diff-1d98f0039bbc6df92adee882ed99d993L78

Hoverbear:
@1000miles I fixed up that issue.

1000miles:
Tested with staging credentials. Works! Thanks a lot! :) 📬

Hoverbear:
Let's wait on merge until we are ready to consciously ready to push this to production since it requires some manual intervention.

Issue #596, opened 2017-03-30
Fix photos uploaded by Wheelmap_Visitor


Svenyo:
Hi @1000miles,

just saw that probably thousands of pics (I assume all pics in the premium pois of Streetspotr) that had been collected and uploaded using the user Wheelmap_visitor from our last Streetspotr crowdsourcing campaigns are not shown in the detailed view.

I assume that we caused this new bug when we recently changed the permission for the Wheelmap_visitor into tagging the accessibility and toilet status only. #477  

Example:
POI "Monkeyman"
https://wheelmap.org/en/nodes/369681660
https://wheelmap.org/en/admin/pois/369681660/photos

![bildschirmfoto 2017-03-30 um 14 05 37](https://cloud.githubusercontent.com/assets/20318534/24503440/6487bcac-1552-11e7-89dc-b50c4c29ac9c.png)

![bildschirmfoto 2017-03-30 um 14 03 59](https://cloud.githubusercontent.com/assets/20318534/24503412/48c34be4-1552-11e7-9803-285a766cd01a.png)

![bildschirmfoto 2017-03-30 um 14 04 53](https://cloud.githubusercontent.com/assets/20318534/24503411/48c320a6-1552-11e7-84d3-fc325a759021.png)

1000miles:
We had a first look into the server logs and this is the **NGINX** output:

```
2017/03/30 14:28:57 [error] 14542#0: *6092606 rewrite or internal redirection cycle while internally redirecting to "/system/uploads/photo/000/067/190/gallery_preview.jpg.html.html.html.html.html.html.html.html.html.html.html", client: <OBFUSCATED>, server: asset0.wheelmap.org, request: "GET /system/uploads/photo/000/067/190/gallery_preview.jpg HTTP/1.1", host: "asset1.wheelmap.org", referrer: "https://wheelmap.org/en/nodes/369681660"
2017/03/30 14:28:57 [error] 14542#0: *6092607 rewrite or internal redirection cycle while internally redirecting to "/system/uploads/photo/000/067/189/gallery_preview.jpg.html.html.html.html.html.html.html.html.html.html.html", client: <OBFUSCATED>, server: asset0.wheelmap.org, request: "GET /system/uploads/photo/000/067/189/gallery_preview.jpg HTTP/1.1", host: "asset1.wheelmap.org", referrer: "https://wheelmap.org/en/nodes/369681660"
```

Similar issue (**internal redirection cycle**) occured recently as reported here:
[...]

PR #599, opened 2017-04-25
Bug/fix streetspotr csv rake tasks

1000miles:
This PR 

**fixes:**
- header rows [:step, :indoor, :toilet, :photo_caption, :photo_url] to match between csv and rake task
- corrects if-else-branching for several cases (e.g. wheelchair_status, etc.)
- mysql2 invalid value encoding error by converting the csv content to ascii before import into the database
- saves previously missing photo_url for `provided_pois` and increments the counter
- activerecord method errors, e.g. `Provider.find_or_create_by_name('Streetspotr')`

**improves:**
- print output to reflect more detailed stats during the import process
- updates streetspotr check rake task

**adds:**
- checks if photos are already imported (will skip if yes)
- stats for saved `photos` and connected `provided_pois`

Related Github issue: #596


Hoverbear:
@1000miles Can you give me what is needed to do a test run of this locally?

1000miles:
Yes, sure. :)

**Check run, e.g.:**
`$ RAILS_ENV=development bundle exec rake streetspotr:check file="DE-spots-20150701-20151104_converted.csv"`

**Import run, e.g.:**
`$ RAILS_ENV=development bundle exec rake streetspotr:import file="DE-spots-20150701-20151104_converted.csv"`

lib/tasks/streetspotr.rake
```diff
@@ -31,77 +35,113 @@ namespace :streetspotr do
...
 
     wheelchair_stati = Hash.new(0)
     toilet_stati = Hash.new(0)
     skipped = Hash.new(0)
+    imported = Hash.new(0)
 
-    CSV.foreach(csv_file, headers: true, header_converters: :symbol, col_sep: ';', row_sep: :auto) do |row|
-      osm_id = row[:refid]
+    # Remove 4-byte characters (e.g. emoji) in strings
+    UTF8_TO_UTF8MB4_CONVERTER = ->(str) { str.encode('utf-8', invalid: :replace, undef: :replace, replace: '').each_char.select { |char| char.bytesize < 4 }.join }
```
Hoverbear: This is pretty clever. :)

Hoverbear: According to [Wiki](https://en.wikipedia.org/wiki/UTF-8) some CJK characters are 4 bytes, for example some kanji like 𠝹 and 𠼮. I believe this code would remove them yes? Is this worth concern?

lib/tasks/streetspotr.rake
```diff
@@ -31,77 +35,113 @@ namespace :streetspotr do
...
         unless poi
-          puts 'Skipped: Invalid Photo Line.'
+          puts "Skipped: No import (osm_id blank and POI not found)!"
+          skipped[:removed] += 1
           next
+        else
+          provided_poi = ProvidedPoi.find_or_initialize_by(poi_id: poi.id, provider_id: provider.id)
+
+          # Find the photo
+          image = Photo.find_by(source_url: row[:photo_url])
```
Hoverbear: We talked about why this was necessary (and why we could not just check hashes or something).

One point was that we reprocess the photos, so the hashes may be different. However I recall that we do save an [original file](https://github.com/sozialhelden/wheelmap/blob/master/app/uploaders/photo_uploader.rb#L139). Perhaps this is a better method?

lib/tasks/streetspotr.rake
```diff
@@ -31,77 +35,113 @@ namespace :streetspotr do
...
+          # Check for photo duplicates
+          if image
+            puts "Skipped: PHOTO #{image.id} already imported."
+            skipped[:removed] += 1
+            next
+          else
+            p = photo(poi, row)
+            p.save!
+            imported[:photo] += 1
+            provided_poi.url = row[:photo_url]
```
Hoverbear: Is this the intended use of this column? https://github.com/sozialhelden/wheelmap/blob/master/db/structure.sql#L373

lib/tasks/streetspotr.rake
```diff
@@ -131,38 +178,69 @@ namespace :streetspotr do
...
   def photo(node, row)
     photo_url = row[:photo_url]
     photo_caption = row[:photo_caption]
     new_photo = node.photos.build
     new_photo.remote_image_url = photo_url
-    new_photo.caption = photo_caption unless photo_caption.blank?
+    new_photo.source_url = photo_url
+
+    unless photo_caption.blank?
+      # Omit photo captions that are longer than 255 varchar
```
Hoverbear: In this case you're not omitting them (that would remove them completely), you're just truncating them.

lib/tasks/streetspotr.rake
1000miles: What do you mean with `Perhaps this is a better method`? I hope I understand you well. :)

Carrierwave will create an `original.jpg` and a few other versions of each photo, no matter if it is uploaded directly via a user or by an external photo_url like streetspotr.

Example (streetspotr):
http://staging.wheelmap.org/nodes/1227709692

**Image location:**
- original version:
  http://staging.wheelmap.org/system/uploads/photo/000/034/198/original.jpg
- gallery_preview version:
http://staging.wheelmap.org/system/uploads/photo/000/034/198/gallery_preview.jpg

mysql database:
```shell
34198 | original.jpg | 1227709692 |       5 | 2015-09-07 15:31:28 | 2017-05-11 14:04:01 |                0 | https://download.streetspotr.com/3afd0ca9f297fd833fccdfae390c94cedf8e121f/3f2c32d3c478041bb0f4ecd1615e9fc4e769e9e7 |

Comment (Hoverbear) [inline lib/tasks/streetspotr.rake]: Right, my question is: Can we see if we already have an image by just checking this `original.jpg`'s SHA hash against the one from Streetspotr?

**Comment by Hoverbear** [comment]
So I ran `bundle exec rake db:migrate` then `RAILS_ENV=development bundle exec rake streetspotr:import file="DE-spots-20150701-20151104_converted.csv"` and got this output on development:

```
// ...
Step: no, Toilet: no, Indoor: yes -> Status: limited, Toilet: no.
rake aborted!
ActiveRecord::RecordInvalid: Gültigkeitsprüfung ist fehlgeschlagen: User muss ausgefüllt werden
/home/ubuntu/bundle/ruby/2.3.0/gems/activerecord-4.1.15/lib/active_record/validations.rb:57:in `save!'
/home/ubuntu/bundle/ruby/2.3.0/gems/activerecord-4.1.15/lib/active_record/attribute_methods/dirty.rb:29:in `save!'
/home/ubuntu/bundle/ruby/2.3.0/gems/activerecord-4.1.15/lib/active_record/transactions.rb:273:in `block in save!'
/home/ubuntu/bundle/ruby/2.3.0/gems/activerecord-4.1.15/lib/active_record/transactions.rb:329:in `block in with_transaction_returning_status'
/home/ubuntu/bundle/ruby/2.3.0/gems/activerecord-4.1.15/lib/active_record/connection_adapters/abstract/database_statements.rb:201:in `block in transaction'
/home/ubuntu/bundle/ruby/2.3.0/gems/activerecord-4.1.15/lib/active_record/connection_adapters/abstract/database_statements.rb:209:in `within_new_transaction'
/home/ubuntu/bundle/ruby/2.3.0/gems/activerecord-4.1.15/lib/active_record/connection_adapters/abstract/database_statements.rb:201:in `transaction'
/home/ubuntu/bundle/ruby/2.3.0/gems/activerecord-4.1.15/lib/active_record/transactions.rb:208:in `transaction'
/home/ubuntu/bundle/ruby/2.3.0/gems/activerecord-4.1.15/lib/active_record/transactions.rb:326:in `with_transaction_returning_status'
/home/ubuntu/bundle/ruby/2.3.0/gems/activerecord-4.1.15/lib/active_record/transactions.rb:273:in `save!'
/home/ubuntu/bundle/ruby/2.3.0/gems/newrelic_rpm-3.15.0.314/lib/new_relic/agent/instrumentation/active_record_helper.rb:34:in `block in save!'
/home/ubuntu/bundle/ruby/2.3.0/gems/newrelic_rpm-3.15.0.314/lib/new_relic/agent.rb:574:in `with_database_metric_name'
/home/ubuntu/bundle/ruby/2.3.0/gems/newrelic_rpm-3.15.0.314/lib/new_relic/agent/instrumentation/active_record_helper.rb:33:in `save!'
/vagrant/lib/tasks/streetspotr.rake:126:in `block (3 levels) in <top (required)>'
/vagrant/lib/tasks/streetspotr.rake:54:in `block (2 levels) in <top (required)>'
Tasks: TOP => streetspotr:import
(See full trace by running task with --trace)
```

I guess I need to create a specific user?

File: lib/tasks/streetspotr.rake (hunk 1 above)
Comment (Hoverbear) [inline lib/tasks/streetspotr.rake]: Please note a description of `"abc123ü𠝹𠼮🐻"` becomes `"abc123ü"` after this process. This means we may lose data on any POIs with CJK characters.

File: lib/tasks/streetspotr.rake (hunk 2 above)
Comment (Hoverbear) [inline lib/tasks/streetspotr.rake]: I created an issue for this in #607 . It might be best to tackle this later.

**Comment by 1000miles** [comment]
> I guess I need to create a specific user with INSERT INTO users (first_name, last_name, email)  VALUES ("Wheelmap", "Visitor", "visitor@wheelmap.org");

Yes, you need to make sure the user `wheelmap_visitor` exists in the database.

File: lib/tasks/streetspotr.rake (hunk 1 above)
Comment (1000miles) [inline lib/tasks/streetspotr.rake]: Thanks for the hint. I'll check this out.

File: lib/tasks/streetspotr.rake (hunk 3 above)
Comment (1000miles) [inline lib/tasks/streetspotr.rake]: Yes, `varchar(255) COLLATE utf8_unicode_ci DEFAULT NULL` is intended for URLs. :)

File: lib/tasks/streetspotr.rake (hunk 1 above)
Comment (Hoverbear) [inline lib/tasks/streetspotr.rake]: We noted that the test string `"abc123ü𠝹𠼮🐻"` cannot be inserted into the database anyways even if the 🐻 is removed

```
Mysql2::Error: Incorrect string value: '\xF0\xA0\x9D\xB9\xF0\xA0...' for column 'caption' at row 1
```

Comment (1000miles) [inline lib/tasks/streetspotr.rake]: We should discuss the solution with @Svenyo and @holgerd at our next meeting if it is necessary/required to also consider chinese, korean and japanese characters encoding. 
https://dev.mysql.com/doc/refman/5.6/en/faqs-cjk.html#faq-cjk-what-cjk-avail

cc @anngreenberg

File: lib/tasks/streetspotr.rake [hunk 5]
```diff
@@ -198,7 +198,7 @@ namespace :streetspotr do
     new_photo.source_url = photo_url
 
     unless photo_caption.blank?
-      # Omit photo captions that are longer than 255 varchar
+      # Truncate photo captions that are longer than 255 varchar
       if photo_caption.length > 255
         new_photo_caption_string = truncate(photo_caption, length: 255, omission: " (...)")
         puts "OMITTED PHOTO CAPTION: #{new_photo_caption_string}"
```
Hoverbear: Please change this `puts` output as well.

lib/tasks/streetspotr.rake
Hoverbear: We noted that is not a problem (yet)

lib/tasks/streetspotr.rake
```diff
@@ -124,13 +125,15 @@ namespace :streetspotr do
...
             p.save!
+            count += 1
             imported[:photo] += 1
             provided_poi.url = row[:photo_url]
             imported[:provided_poi] += 1
             puts "Success: PHOTO for poi_id #{poi.id} saved!"
           end
 
           provided_poi.save!
+          count += 1
```
Hoverbear: So if `image` evaluates to `false` then we increment `count` twice?

1000miles: Yes, at the moment the counter works a bit different. We use the counter to have a control system on ALL `save` actions which means it does not check the logic/relations behind, e.g. if it is an update action actually, etc. This way we can compare between:

- `imported[:photos]`
- `imported[:provided_pois]`
- total saved actions 

There is probably a way to have a more precise incremental/decremental counter by using `:counter_cache` and  invoking callbacks. Since this requires a new column in the database and migration to enable counter_cache on pois, I would suggest to do this Github issue #612. What do you think?

```ruby
count = 0
...

poi = Poi.find_by(osm_id: osm_id)

if poi
 ...
  provided_pois.save!
  count += 1 # => 1
else
  provided_pois.save!
  count += 1 # => 1
end

provided_pois.save!
count += 1 # => 2

# Counter: 2 (saved actions)
```

Referencing some links:
- http://guides.rubyonrails.org/association_basics.html
- https://stackoverflow.com/questions/26019392/rails-4-1-counter-cache-not-updating
- https://gist.github.com/leemour/ea15148bc95ea31f3a1a

Hoverbear: So the counter just counts the number of times `save` is called? Can you please rename the variable to `save_call_count` or document it somehow?

lib/tasks/streetspotr.rake
```diff
@@ -31,77 +35,117 @@ namespace :streetspotr do
 
   desc 'Import data from StreetSpotr'
   task import: :environment do
+
     csv_file = ENV['file']
     raise 'Usage: bundle exec rake streetspotr:import file=<your_csv_file>' unless csv_file
 
     poi = nil
```
Hoverbear: Please note

> This is due the the case where the next line may not have an attached OSM ID but is a photo for the previously seen OSM ID.

So when we look at this in a few months we're not 😕 .

lib/tasks/streetspotr.rake
```diff
@@ -142,6 +145,7 @@ namespace :streetspotr do
     puts "Toilet: Yes: #{toilet_stati[:yes]}, No: #{toilet_stati[:no]}, Unknown #{toilet_stati[:unknown]}."
     puts "Newly Saved: Photos: #{imported[:photo]}, Newly Connected: ProvidedPoi: #{imported[:provided_poi]}."
     puts "Skipped: Unknown: #{skipped[:unknown]}, Skipped: Not imported: #{skipped[:removed]}."
+    puts "Successful Saved Actions: #{count} counted."
```
Hoverbear: `puts "Successful save! calls: #{count}."` may be more appropriate to describe what's happening.

1000miles:
@emig Can you please review this PR? Thank you. :)

lib/tasks/streetspotr.rake
```diff
@@ -1,168 +1,172 @@
...
+    puts "EXISTING: ProvidedPois: #{pp.count}, Photos: #{photos.count}."
+    puts "SKIPPED: Photos: #{@skipped[:photo]}, Provided Pois: #{@skipped[:provided_poi]}."
+    puts "SAVE ACTIONS: Photos: #{@saved[:photo]}, Provided Pois: #{@saved[:provided_poi]}"
+    puts "TOTAL SUCCESSFUL SAVE ACTIONS: #{@count}."
+
+    previous_poi = current_poi
   end
 
-  def photo(node, row)
+  def photo(poi, row)
```
emig: here we can just pass the photo_url string, no need to pass the `row` instance, since we won't use it in the function

lib/tasks/streetspotr.rake
```diff
@@ -1,168 +1,172 @@
...
+    new_photo.source_url = photo_url
     new_photo.user = User.wheelmap_visitor
     new_photo
   end
 
-  def minimal_status(stati)
-    return 'no' if stati.include?('no')
-    return 'limited' if stati.include?('limited')
-    stati.first
+  def photo_check(poi, row)
```
emig: here we can just pass the photo_url string, no need to pass the row instance, since we won't use it in the function

lib/tasks/streetspotr.rake
```diff
@@ -1,168 +1,172 @@
...
+    else
+      p = photo(poi, row)
+      p.save!
+      @count += 1
+      @saved[:photo] += 1
+      puts "Success: PHOTO for osm_id #{poi.id} saved!"
+    end
+  end
+
+  def photo_check_dryrun(row)
```
emig: here we can just pass the photo_url string, no need to pass the row instance, since we won't use it in the function

lib/tasks/streetspotr.rake
```diff
@@ -1,168 +1,172 @@
...
-        if status == 'unknown'
-          puts 'Skipped: Unknown Status.'
-          skipped[:unknown] += 1
+        unless current_poi
+          puts "Skipped: POI for osm_id #{osm_id} not found."
+          @skipped[:provided_poi] += 1
           next
+        else
+          find_or_initialize_provided_poi(current_poi, provider)
+          photo_check(current_poi,row)
```
emig: here we can just pass the photo_url string, no need to pass the row instance, since we won't use it in the function

lib/tasks/streetspotr.rake
```diff
@@ -1,168 +1,172 @@
...
+    provided_poi.save!
+    @count += 1
+    @saved[:provided_poi] += 1
+    puts "Success: Provided Poi with provided_poi_id #{provided_poi.id} saved!"
   end
 
-  def wheelchair_toilet(status, toilet)
-    return 'unknown' if status == 'unknown'
-    toilet
+  def find_provided_poi(poi, provider)
```
emig: here we can just pass the poi.id and provider.id integers, no need to pass the poi,provider instances, since we won't use it in the function

lib/tasks/streetspotr.rake
```diff
@@ -1,168 +1,172 @@
...
+      @saved[:photo] += 1
+      @count += 1
+    end
   end
 
-  def wheelchair_status(has_step, _indoor)
-    # TODO: Add the indoor again when the Streetspotr questions got fixed
-    return 'no' if has_step == 'yes'
-    'unknown'
+  def find_or_initialize_provided_poi(poi, provider)
```
emig: here we can just pass the poi.id and provider.id integers, no need to pass the poi,provider instances, since we won't use it in the function

lib/tasks/streetspotr.rake
```diff
@@ -1,168 +1,172 @@
...
-          next
-        end
-
-        step = has_step(row)
-        toilet = toilet(row)
-        indoor = indoor(row)
-
-        status = wheelchair_status(step, indoor)
-        toilet = wheelchair_toilet(status, toilet)
+        current_poi = Poi.find_by(osm_id: osm_id)
```
emig: The method `find_by!` would raise  `ActiveRecord::RecordNotFound`. We can make use of that to
simplify the branching. (note the  `!`)

You can put the code from lines -99 on a rescue block for that exception, and you save
a whole unless/else branching

lib/tasks/streetspotr.rake
1000miles: Discussed and agreed with @emig:

Since we are using `poi` [here](https://github.com/sozialhelden/wheelmap/blob/1251e00224aca74b3944bf6bd4b4978ffb1ebbf6/lib/tasks/streetspotr.rake#L119)
and the method `photo_check` relies on the method `photo`, we can not remove it.

emig:
Since both rake tasks are doing the same work, isn't it possible to define the check task in terms of the other task, passing an extra flag --dry-run, for example? There is a lot of duplicated code

1000miles:
Thank you. 👯 :D

1000miles:
Deployed to production today with release [20170626103531](https://github.com/sozialhelden/wheelmap/releases/tag/production-20170626103531).

PR #608, opened 2017-05-12
[DONT MERGE YET] fix:  misspelling typo

### LINKED ISSUE #598: `typ` mispelled on node detail page.
On https://wheelmap.org/nodes/1976439634 you can see that `typ` is spelled incorrectly:

screen shot 2017-04-24 at 11 57 50

emig:
- recompiles i18n-js generated files

**note**: 

@Hoverbear couldn't get the link on local machine, is it because locally we are missing the region, and therefore we don't have that feature activated? Therefore I couldn't test this fix works ...

screen shot 2017-05-12 at 10 33 43

closes #598

### DIFF
--- a/app/assets/javascripts/i18n/en.js
+++ b/app/assets/javascripts/i18n/en.js
@@ -1030,7 +1030,7 @@ I18n.translations["en"] = I18n.extend((I18n.translations["en"] || {}), {
       "link": {
         "back": "back",
         "large_map": "Enlarge",
-        "listing": "All places of typ '%{type}' in %{city}",
+        "listing": "All places of type '%{type}' in %{city}",
         "upload": "Upload image"
       },
       "more_data_from": "There is more information about this place, validated by:",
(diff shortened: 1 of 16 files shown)


Hoverbear:
You can test this by either removing the exclude filter, or increasing the accuracy to like 100000 in `app/views/nodes/_node_nearby.html.haml`

emig:
Hum, didn't work, 
seems the node in question doesn't have a region:

https://github.com/sozialhelden/wheelmap/blob/598_mispelling/app/views/nodes/_node_similar.html.haml#L13

emig:
But one is about similar nodes and the other about nearby ones, seem conceptually unrelated.

emig:
@Hoverbear none of the pois on the local db has `region` set, so the line I mentioned before won't be active for any poi. https://github.com/sozialhelden/wheelmap/blob/598_mispelling/app/views/nodes/_node_similar.html.haml#L13

Hoverbear:
@emig I get results doing this.

emig:
wasn't ware of the import region rake taks, thanks!

emig:
@Hoverbear on a fresh install of the project via vagrant the imported data doesn't contain regions. You can test this then, if you have them.

Hoverbear:
foo

emig:
We are looking for a link like `All places of type ... in Berlin`, do you have that? It depends on the POI having a region

Hoverbear:
Ah, my mistake. I don't see that.

PR #618, opened 2017-05-18
feat: keep vm clock update via ntp

emig:
installs ntp preventing problems with ssh certificates not being active since a present date, when
vm has its clock set in the past


Hoverbear:
Does this start the service as well?

emig:
Yes

Xylakant:
keep in mind that there's a limit on how much drift the ntp daemon will correct. suspend/resume may still not work as expected.

PR #632, opened 2017-06-26
Add node_type to short node GeoJSON

opyh:
This allows the new frontend to use the node_type attribute for map markers.

Background: The new frontend includes a mechanism to categorize PoIs using synonymes. When developing accessibility.cloud, Holger, Thomas, Katharina and I unified the various systems we found. The new frontend displays subtly different icons for that reason, and to do this correctly, it needs the exact node_type info for each PoI.

Tested locally. Could somebody merge / deploy this? This is hopefully the last step for displaying markers from different sources on a map correctly in the new frontend 🤞🏻

### DIFF
--- a/app/models/poi.rb
+++ b/app/models/poi.rb
@@ -323,6 +323,7 @@ class Poi < ActiveRecord::Base
                     'wheelchair_toilet' => wheelchair_toilet,
                     'id'         => osm_id,
                     'category'   => category.try(:identifier) || '',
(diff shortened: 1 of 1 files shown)


1000miles:
@opyh 

Deployed to staging today with release [20170627124206](https://github.com/sozialhelden/wheelmap/releases/tag/staging-20170627124206).

opyh:
Tested, works for me, thank you!

Issue #633, opened 2017-06-29
[CSV Import] 'ArgumentError: argument out of range' on date_time


1000miles:
This is a follow-up ticket for https://github.com/sozialhelden/wheelmap/issues/630#issuecomment-311650076.

### Background:
At the moment we can not fully loop through the files `DE_2016.csv`, `IR_2016.csv` and `UK_2016.csv` to import the photo images from streetspotr.

Following are the conditions to debug in dev:
- the osm_id in the csv must match with the osm_id of the existing poi in the dev environment (local)
- the poi/record must raise the exception `(ArgumentError): argument out of range`

In order to analyze what is wrong with the date time on a specific record we need to be able to compare the EXIF data of an image - that requires to strip and display the exif data of respective image before hand.

Needs more investigation.

**Maybe related:**
- https://stackoverflow.com/questions/9131697/rails-active-record-argumenterror-argument-out-of-range-for-time-fields-g
- https://stackoverflow.com/questions/4076248/mysql-time-type-gives-an-argumenterror-argument-out-of-range-in-rails-if-ov
- https://stackoverflow.com/questions/27115441/random-intermittent-argument-out-of-range-error-while-parsing-csv-and-parsing

1000miles:
**Update, 2017 July 5:**

We finally found out why some images cause an `(ArgumentError): argument out of range` error related to datetime. 

Following 3 images are examples for comparison:
- https://download.streetspotr.com/0da9ec28c99f1c1a598da20fac6432c6dd8cf08e/a9cf07fa864fb2d6ee2d01fe5d6665d80f147e82 (working)
- https://download.streetspotr.com/81e38f251e3a6bde77ce7b027256ec27276429ca/a6d74cb779fc4e12b14ab0152c46fb9e14ebf176 (working)
- https://download.streetspotr.com/05d67819ca0c286c542e88c322a58a4a9115b2d4/1b6fb57f65ae00c36e6933ab2b23c412b79c45bd ((ArgumentError): argument out of range)

```shell

# working
>> EXIFR::JPEG.new('photo_2517845_original.jpg').date_time
=> 2017-05-08 12:07:45 +0200
>> EXIFR::JPEG.new('photo_2075939_original.jpg').date_time
=> 2017-01-21 12:28:43 +0100

# not working
>> EXIFR::JPEG.new('photo_2457067_original.jpg').date_time
=> "2017:04:24 5:37:02 PM"
```

anngreenberg:
[...]

Issue #630, opened 2017-06-15
Importing photos from CSV files sent by Streetspotr


Svenyo:
So, here are the files with photos from Streetspotr from the current campaign.

There is one file for each country: Belgium (BE), Germany (DE), Denmark (DK), Spain (ES), France (FR), Italy (IT), Ireland (IR), Netherland (NL), United Kingdom (UK).

Please import them into the database.
[Streetspotr CSV files 062017.zip](https://github.com/sozialhelden/wheelmap/files/1077788/Streetspotr.CSV.files.062017.zip)

Svenyo:
@1000miles @holgerd @anngreenberg If this is too short in notice for the current milestone list please consider this issue to be important for the next milestones starting on June, 19.

1000miles:
Thank you for the files, @Svenyo. We will test your new CSV files today and try to deploy them to staging. As soon as this is done, we will inform you to review them on staging. If everything works well, we can deploy them to production on monday (2017, june 19) after final approval.

1000miles:
Copied comment from #167:

@Svenyo cc @holgerd 

> IMPORTANT: We need to delete the broken images first before we can do new imports from csv via the rake task, otherwise the photo duplicate checker will not import the photos because it thinks the photo already exists (even though it is broken). We will wait for your check and final approval to remove the corrupt 14733 effected images before we import new csv data from streetspotr.

1000miles:
@Svenyo The files inside the zip file `Streetspotr CSV files 062017.zip` are all named `BE_2016.csv`, DE_2016.csv`, etc. Can you please confirm if `2016` data is correct?

1000miles:
### Update, 2017, 19 June:

Following data exist on staging at the moment:
- pois: 1.004.426
- provided_pois: 3.214
- photos: 25.819

**General rule: We can only import data from csv if the poi already exists in the database.**

Even though these data listed above on staging seem to be a lot it looks as if we do not have enough data on staging to do the test import. I did a dry-run  with `DE_2016.csv` that contains 19197 records. 

Here is an example output:

```bash
RAILS_ENV=staging bundle exec rake streetspotr:check file="DE_2016.csv"

POIS not found: 0.
POIS and OSM_ID nil: 19196.
CREATED: Photos: 0
CREATED: Provided Pois: 0
EXISTING BEFORE IMPORT: ProvidedPois: 3214, Photos: 25819.
[...]

PR #634, opened 2017-07-06
Fix/argument error on datetime

1000miles:
At the moment we only fetch the EXIF date_time of an image (before photo creation) but we do not validate the format of the date_time. This PR specifies 3 ways of date_time, anything beyond them will set the EXIF date_time to `nil`.

- '2017:04:22 03:00:22 +0200'
- '2017-06-08 13:40:32 +0200'
- '2017/06/08 1:40:32 +0200'

Fixes Github issue: #633.
Github issue #630 is dependent on this PR.


1000miles:
Thaaaank you, @justahero ! 🍦 🍬 We will test and deploy this on monday. :)

Issue #606, opened 2017-05-08
wheelchair:description does not show up on wheelmap


anngreenberg:
As reported by @ldodds on issue #600  

> A user doing mapping using the app added a comment to a coffee shop. While the comments were saved to OSM, they are not appearing in the mobile or web versions of Wheelmap.
> 
> OSM: http://www.openstreetmap.org/way/452678133
> Wheelmap: https://wheelmap.org/en/nodes/-452678133
> 
> Let me know if I should file this as a separate report. Am a bit concerned that notes will get overwritten or lost if they're not displaying

From a quick check, this is also true for other nodes, as for example 
OSM: http://www.openstreetmap.org/way/462307362
Wheelmap: https://wheelmap.org/nodes/-462307362

When adding a comment, this should also be displayed on wheelmap directly.

holgerd:
Just tested the following:

- I was logged in on wheelmap.org and edited the wheelchair:description of this place: https://wheelmap.org/en/nodes/1569218581 (from "Hat eine Rampe" to "has a ramp")
- I see the change on www.openstreetmap.org/node/1569218581 and expect to see it on wheelmap as well
- Instead, I stell see the old description on https://wheelmap.org/en/nodes/1569218581 5 minutes after editing it.

holgerd:
It is okay now! https://wheelmap.org/en/nodes/1569218581

holgerd:
Just tested again and it works:

- I used the iPhone app to add a comment to Cafe Sur in Berlin
- I open Cafe Sur in the web browser: https://wheelmap.org/nodes/623350251 and I see the comment (after ca 3 minutes)
- I open the detail page on the iPhone and do not see the comment. I tap on "edit comment" and there I see my comment in the textarea. This seems to be the only view of comments in the iphone app.

holgerd:
Another test, this time with a node with negative ID:

- I used the iPhone to add the comment "test" to this church: https://wheelmap.org/nodes/-23453689
- This is the changeset: http://www.openstreetmap.org/changeset/49691034
- On iOS I can see the comment when I open the "edit comment" again
- **The comment does NOT show up** on https://wheelmap.org/nodes/-23453689 even 26min later

holgerd:
Just tested:
The comment for https://wheelmap.org/nodes/-23453689 neither shows up on the detail page of the Android app.

1000miles:
### Update, 2017-07-18

For now we identified 3 behaviors:

[...]

PR #642, opened 2017-08-02
Fix/wheelmap description missing on nodes

1000miles:
This PR fixes **behavior one** listed in issue (https://github.com/sozialhelden/wheelmap/issues/606#issuecomment-315998907) :
- Some text being posted in the comment/description section of affected POIs are not saved to the wheelmap database and therefore do not show up on a wheelmap POI detail page even though they are visible on osm. 

Fix:
- in the `update_poi!` method we previously only saved the tags `wheelchair` and `toilets:wheelchair` but not `wheelchair:description` until now.


1000miles:
Thanks for the review, @justahero! 👯

Issue #679, opened 2018-05-23
Add link to privacy policy to footer (for GDPR compliance)


holgerd:
In the lower right corner of https://wheelmap.org/ is a text link "Impressum" / "Imprint". This leads to https://news.wheelmap.org/impressum/ if the language is German and to https://news.wheelmap.org/en/imprint/ in any other language.

To the right side of this text link we need another textlink following the same logic.

Text: 

- "Datenschutzbedingungen" in German and 
- "Privacy Policy" in any other language (please discuss translation implications with @Svenyo )

Link:
 
- in German: https://news.wheelmap.org/datenschutzbestimmungen/
- in any other language: https://news.wheelmap.org/en/privacy/

Issue #669, opened 2018-02-15
Bring Mongolian language version online


Svenyo:
In Transifex, the language file for Mongolian has been completed. Please implement it from the project https://www.transifex.com/sozialhelden/wheelmap-website/language/mn/ to http://staging.wheelmap.org/map so the translator can review the translations in the Wheelmap environment.

- The URL should be:  //wheelmap.org/mn/map
- The Keyword for the dropdown menu is: монгол хэл

Please let me know if there are further questions. Thanks!

spacekookie:
@Svenyo The mongolian translations have been deployed to https://staging.wheelmap.org/mn/map

Please have a look if everything is ok with them. We're still waiting on the translations of the new terms ("Datenschutzerklärung"/"Privacy Policy") which currently default to German. Are you ok with the current version being deployed to production before that?

PR #681, opened 2018-05-23
Adding links about GDPR compliance to the footer

spacekookie:
This is addressing issue #679 but not done yet. We need to figure out how to handle the translations for the newly added link. Right now no text is present (not even German/ English)

**Note** This PR also adds translations for #669

### DIFF
--- a/app/assets/javascripts/i18n/ar.js
+++ b/app/assets/javascripts/i18n/ar.js
@@ -774,6 +774,7 @@ I18n.translations["ar"] = I18n.extend((I18n.translations["ar"] || {}), {
       "map": "خريطة",
       "newsletter": "أخبار الموقع",
       "press": "صحافة",
+      "privacy": "Privacy Policy",
       "profile": "الحساب",
       "projects": "قم بالاشتراك",
       "report_problem": "بلغ عن مشكلة",
--- a/app/assets/javascripts/i18n/bg.js
+++ b/app/assets/javascripts/i18n/bg.js
@@ -774,6 +774,7 @@ I18n.translations["bg"] = I18n.extend((I18n.translations["bg"] || {}), {
       "map": "Карта",
       "newsletter": "Newsletter",
       "press": "Преса",
+      "privacy": "Privacy Policy",
       "profile": "Профил",
       "projects": "Включете се",
       "report_problem": "Съобщи за проблем",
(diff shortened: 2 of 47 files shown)


holgerd:
@spacekookie Thanks for this. I do not understand: Do you need text from us? cc @Svenyo

spacekookie:
Your original issue said to discuss translations with you. We can of course only add the German and English values ourselves, but then I'm not sure what users visiting for another language would encounter.

Svenyo:
Hi @spacekookie, please find the general workflow for translations here: https://github.com/sozialhelden/wheelmap/wiki/Add-new-translations 
What's most important is that...

1. new text strings need to be integrated in the source language (English) of the corresponding .yml file and 
2. uploaded again in Transifex. 
3. When this is done, please send me a short message, 
4. then I can start the procedure for getting the translations done with Transifex. 
5. I'll send you a message, when translations are completed and the .yml files are ready for use.

spacekookie:
Ping @Svenyo, you can start the translation process now

Svenyo:
Hi @spacekookie, I can't find the string in Transifex. Is the updated source language (English) already uploaded?

spacekookie:
Hey @Svenyo, I added the modified `.yml` files for transifex in this pull request:

 - German: https://github.com/spacekookie/wheelmap/blob/5081f719270e239430afc0a0b4bdd8d644115e36/config/locales/de/relaunch.yml
 - English: https://github.com/spacekookie/wheelmap/blob/5081f719270e239430afc0a0b4bdd8d644115e36/config/locales/en/relaunch.yml

From what I understand these need to be provided to transifex so we can download the translated files for wheelmap.

spacekookie:
According to [this wiki article](https://github.com/sozialhelden/wheelmap/wiki/Translation-Guide) I should just upload the `.yml` files with the new keys. Then:

> 2. The updated resource files are to be uploaded to Transifex by the product manager

So I was under the assumption that you had to do that.

Svenyo:
Okay, good!

Svenyo:
Yes, good idea. Is it possible to deploy it to staging first for the review before it goes to production?

spacekookie:
@Svenyo Alright, we're merging this PR now then and deploying to staging

Svenyo:
Dear @spacekookie, just checked on staging: For the German version everything looks fine on staging, for the English version as well. When choosing all other languages, the English wording and privacy policy content should appear, though. Please adjust!

1000miles:
@Svenyo cc @spacekookie 

What you currently see on staging in the footer at the moment ("Datenbestimmungen" on all pages except for English locale) is expected behavior. :) `:de` is the default locale in the wheelmap project while `:en` is the default locale for translations which explains why the **"Datenschutzbestimmungen"** appears on all other locale pages that don't have the translations yet.

PR #683, opened 2018-06-01
Fix: Upgrade and pin jquery-ui-rails ~> 5.0.5

1000miles:
During deployment of `Removing airbrake` (https://github.com/sozialhelden/wheelmap/pull/682) to **staging** we encountered following assets precompile error:

```
rake aborted!
Sprockets::FileNotFound: couldn't find file 'jquery-ui/datepicker' with type 'application/javascript'
```

Link to solution: https://github.com/activeadmin/activeadmin/issues/4672

This PR upgrades and pins `jquery-ui-rails ~> 5.0.5` to ensure assets precompiling works again.

PR #682, opened 2018-05-30
Removing airbrake

spacekookie:
This PR replaces #678

PR #685, opened 2018-06-05
[Fix] Pin Capistrano '~>3.4' & Capistrano-Deploytags

1000miles:
When we deployed PR #682 & #683 to staging everything worked as expected but when we deployed  these PRs to `production` with capistrano it suddenly required a ssh passphrase. Even when we typed in the correct password it did not accept it which let the deployment being stuck at that point. 

In debugging mode we got following error:

```
Net::SSH::AuthenticationFailed: Authentication failed for user vagrant@xx.x.xx.x...
```

**Background info:**
Since Capistrano version '> 3.2.0' a ssh password feature is provided for deployment. As long as we don't adjust the deployment settings we would need to downgrade back to the previous capistrano version that is BEFORE '~>3.7' (otherwise it would not work together with capistrano-deploymentags).

More info: http://capistranorb.com/documentation/faq/how-can-i-get-capistrano-to-prompt-for-a-password/

This PR does following:
- downgrades to Capistrano '~>3.4.x'
- downgrades to Capistrano-Deploytags to '~> 1.0', '>= 1.0.6'
- upgrades to Bundler v1.16.2

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -124,10 +124,10 @@ group :production, :vagrant do
 end
 
 group :development do
-  gem 'capistrano', '>= 3.1.0'
-  gem 'capistrano-bundler',                 require: false
-  gem 'capistrano-deploytags', '~> 1.0.0',  require: false
-  gem 'capistrano-npm',                     require: false
-  gem 'capistrano-rails',                   require: false
-  gem 'capistrano-rbenv',                   require: false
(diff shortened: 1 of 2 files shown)


1000miles:
Deployed to production today with release https://github.com/sozialhelden/wheelmap/releases/tag/production-20180605101432.

PR #687, opened 2018-06-22
Adding mongolian language configs and various translation improvements

spacekookie:
This addresses issue #669.

@Svenyo Please upload [these files](https://github.com/sozialhelden/wheelmap/tree/a836eb818c61c3c8ebd4e3e676384fcb6c48f569/config/locales/mn) to transifex so that they can get started on the translation process


spacekookie:
Okay, after checking the issue again I saw that the files were already translated into mongolian. Sorry about that!

I pulled the translations and generated the assets. Will test this and can then deploy it to staging.

**Note** This doesn't include any of the translations for the recent GDPR update yet, I guess we will have to update the translations again once they become available from transifex?

Issue #686, opened 2018-06-20
Toolbar in English if translations not present


spacekookie:
Normally the website defaults to being in German if for a selected language a term is not available.

The toolbar doesn't follow that rule and is then always in English. 

There should be more consistent behaviour in this: either have it all in English or all in German if translation terms can't be found for the selected language.

Possibly related files (add more):

- `app/views/relaunch/_header.html.haml`

PR #688, opened 2018-06-25
Switching the default locale to :en

spacekookie:
Addresses an issue discovered in #681 (related #686)

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -29,10 +29,10 @@ module Wheelmap
     config.middleware.use(Rack::I18nLocaleSwitcher)
     config.i18n.fallbacks = true
     config.i18n.load_path += Dir[Rails.root.join('config', 'locales', '**', '*.{rb,yml}')]
-    config.i18n.default_locale = :de
-    config.i18n.enforce_available_locales = false
+    config.i18n.default_locale = :en # We changed the default locale to english because #681
+    config.i18n.enforce_available_locales = true
 
-    config.i18n.available_locales = [:de]
+    config.i18n.available_locales = [:en]
     # Use all locales that we have translations when not in production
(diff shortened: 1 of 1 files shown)


spacekookie:
@1000miles Any idea about this failure?

1000miles:
@spacekookie cc @Svenyo 

Please keep `:de` as default locale. 

2. A lot of test cases are based on german locale, if you change it to `:en` you would need to adjust/fix all the test cases as well. I assume that's why your PR https://github.com/sozialhelden/wheelmap/pull/688 fails.

3. In the past we also wanted to unify the default locale for the wheelmap project AND transifex but as far as I remember @holgerd once said they want to keep the german locale because of historical reasons, e.g. search engines already have `:de` indexed as default locale.  

4. If we introduce new keys & translations in english without providing the german translations as well, the system would throw an error message like `translation key is missing` anywhere where no translations in other locales exist yet. 

**My suggestion:** Keep it as it is and wait until new translations are coming up for "Datenschutzbestimmungen". 

**Note:** Even though it works at the moment I think for consistency the key `privacy` should be put to the keys structure for `footer` instead of `header`. Here is an example where the structure maps to the translations resource files: 

https://github.com/sozialhelden/wheelmap/blob/8df541107ed785f82a2436ebe53d11129e7e2fba/config/locales/en/wheelmap.yml#L351-L353

=> https://github.com/sozialhelden/wheelmap/blob/master/app/views/relaunch/_footer.html.haml#L30

spacekookie:
@1000miles  That would be fine for me. Depends on what @Svenyo thinks.

Also, no I didn't run the tests locally, I was moving from place to place a lot. But…checking tests asynchronously is kinda what failing PR CI is for, isn't it :wink:

